PR #891, opened 2026-04-29
add scheduled task to disable expired api keys

MorganaFuture:
## Description
Expired API keys stayed enabled=true in the DB — gateway rejected them at request time, but stored state was stale.

## Improvements
- Added ApiKeyExpirationScheduler - hourly sweep with ShedLock, configurable via openframe.api-key-expiration.*
- Added ApiKeyExpirationService that disables expired keys via existing findExpiredKeys() query
- Removed dead TODO method from ApiKeyService
- Unit tests for both the bug and the fix


MorganaFuture:
@ivan-flamingo could you please take a look?

PR #941, opened 2026-05-14
fix(InfoCard): adjust value row gap to 4px mobile / 8px desktop

RaminMakhmudov0990:
## Improvements
- 
-

### DIFF
--- a/openframe-frontend-core/src/components/ui/info-card.tsx
+++ b/openframe-frontend-core/src/components/ui/info-card.tsx
@@ -98,7 +98,7 @@ function InfoCardValueRow({ label, value, showLabel, copyable, copyAriaLabel }: 
   const { copy, copied } = useCopyToClipboard()
 
   return (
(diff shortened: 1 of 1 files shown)


openframe-frontend-core/src/components/ui/info-card.tsx
```diff
@@ -98,7 +98,7 @@ function InfoCardValueRow({ label, value, showLabel, copyable, copyAriaLabel }:
   const { copy, copied } = useCopyToClipboard()
 
   return (
-    <div className="flex h-6 items-center gap-0.5 self-stretch w-full">
+    <div className="flex h-6 items-center gap-1 md:gap-2 self-stretch w-full">
```
oleksandr-blip: responsive `gap-[var(--spacing-system-xs)]` replaces both `gap-1` and `md:gap-2`

oleksandr-blip (changes requested):

oleksandr-blip (approved):

PR #1173, opened 2026-06-15
Script API: Integrate Back-End with Front-End

andrii-flamingo:
Closes https://app.clickup.com/t/9013925967/86aj1ux3n

## Summary by CodeRabbit

* **Documentation**
  * Added detailed GraphQL contract docs for saved scripts, including CRUD, run command dispatch, and execution cancellation behavior.
* **API Updates**
  * Script updates now require the script id in `UpdateScriptInput` (simplified mutation shape).
  * Script deletion now returns the deleted script id, including idempotent no-op behavior.
  * Script `status` is now a strongly typed enum.
  * Script listings now expose `filteredCount` (tenant-filtered totals independent of pagination).
* **Tests**
  * Updated and added coverage for the new update/delete/list counting behaviors.


openframe-api-lib/src/main/java/com/openframe/api/mapper/ScriptMapper.java
```diff
@@ -61,7 +62,7 @@ public ScriptResponse toResponse(Script entity) {
                 .defaultTimeoutSeconds(entity.getDefaultTimeoutSeconds())
                 .defaultArgs(entity.getDefaultArgs())
                 .envVars(mapEnvVarsToResponse(entity.getEnvVars()))
-                .status(entity.getStatus() != null ? entity.getStatus().name() : null)
+                .status(entity.getStatus() != null ? entity.getStatus().name() : ScriptStatus.ACTIVE.name())
```
kirill-567, line 65: What's the reason status null?
it should be non null always
Just active -> archived -> removed

kirill-567 (approved):

openframe-api-lib/src/main/java/com/openframe/api/mapper/ScriptMapper.java
andrii-flamingo: Fixed

PR #1408, opened 2026-07-09
fix(cards-strip): reliable hover on moving track — pointer-tracked re-sync

michaelassraf:
## Problem
Hover play/pause on the strips was intermittent: browsers do NOT re-dispatch pointerenter/leave when CONTENT moves under a STATIONARY pointer (marquee drift, chevron glide, seam warp, native scroll). Cards slid under a still cursor without activating, or slid/warped away leaving a stale hovered card playing.

## Fix (2026 pointer best practices)
- Track the last hover-capable pointer position (`pointermove` on the scroller, `pointerType === 'touch'` excluded) and re-resolve the hovered card via `document.elementFromPoint` + `data-strip-card-key` markers **whenever the track moves** (rAF tick + every scroll event) — hover activation now always means "the pointer is over THIS card's area": gaps/whitespace deactivate, per-card pointer handlers remain the zero-latency fast path.
- Scroller `pointerleave` force-deactivates (covers leave events swallowed by DOM churn under the cursor).
- Markers: card root in render-prop mode (`VideoBiteCard`), managed cell in children mode — inert clone content resolves to its non-inert cell, shadow-DOM players resolve to their host.
- Also: chevrons now use the shared `Button variant="outline" size="icon"` (follow-up to #1405's engine extraction).

## Verified live
Real-browser test on the flamingo homepage strips: stationary pointer + moving track → pauses and activates exactly the card under the cursor; leave → resumes at 60px/s; gap hover deactivates.

Follow-up to #1405 (merged as 0.0.395) — needs a release + hub pin bump to reach previews/prod.

## Summary by CodeRabbit

* **New Features**
  * Added a new portrait-style card layout across multiple content cards, giving more consistent image, author, and link presentation.
  * Expanded program and content card support for portrait density in more places.

* **Bug Fixes**
  * Improved hover behavior in card strips so highlighted cards stay in sync as content moves.
  * Added better image fallback handling for portrait cards when images are missing or poorly sized.

* **Refactor**
  * Updated card strip rendering to support mixed card widths and smoother scrolling interactions.


openframe-frontend-core/src/components/chat/entity-cards/case-study-card.tsx
```diff
@@ -48,7 +49,7 @@ export interface CaseStudyCardProps {
   targetPlatform?: string | null
   /** OG placeholder URL, used when `study.featured_image` is missing. */
   placeholderUrl?: string | null
-  size?: 'default' | 'sm'
+  size?: 'default' | 'sm' | 'portrait'
```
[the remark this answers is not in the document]
michaelassraf: Fixed in 76026d06 — skeleton size union widened to `'default' | 'sm' | 'portrait'` (portrait shares the default skeleton shape since the portrait anatomy uses the same zone boxes), and removed a stray `| 'sm' | 'portrait'` fragment that had leaked onto `showTypeBadge`.

openframe-frontend-core/src/components/chat/entity-cards/entity-portrait-card.tsx
```diff
@@ -0,0 +1,205 @@
...
+  placeholderUrl?: string | null
+  imageAlt: string
+  title: string
+  /** Typography override for the title (entity identity), zone box unchanged. */
+  titleClassName?: string
+  person?: EntityPortraitPerson | null
+  className?: string
+}
+
+export function EntityPortraitCard({
```
michaelassraf, line 74: we should have common logic for fallback, why do we have this here?

michaelassraf: You're right — extracted in 9ee1450b. NEW `use-cover-image-fallback.ts` (next to the other entity-card hooks) is now THE one cover → placeholder → hide chain: `EntityPortraitCard` and `BlogCard` both use it (BlogCard's hand-rolled `imageError`/`displayImage` copy is deleted), and the copy-pasted `hideOnError` in case-study + interview cards is deduped into the same module. Bonus over the old copies: a failing *placeholder* now hides instead of staying broken, and a changed cover URL re-resolves with no reset effect (state = set of failed URLs, src derived).

PR #1442, opened 2026-07-14
feat(notifications): OS push delivery via FCM — device tokens + broadcaster sink

semen-flamingo:
Backend for **Native OS Notifications — Push Infra** (ClickUp `86ajebb5c`). Extends `NotificationBroadcaster`, until now socket-only, with a push sink, plus the device-token store push depends on.

## Design

**One provider, not two.** iOS hands us an FCM token too (the client uses the Firebase messaging plugin) and Firebase forwards to APNs itself — so there is no per-platform routing and no APNs `environment` to track. `PushDevice.environment` is kept nullable only so a direct-APNs provider stays possible without a migration.

**`PushSender` port lives in `openframe-data-nats`, the implementation does not.** The port deliberately mirrors `publishToUser(userId, notification, category)`, so the broadcaster treats push as just another per-recipient sink and stays dumb. `firebase-admin` (heavy: google-cloud + gRPC) is confined to the new opt-in `openframe-notification-push` module — it must never leak into `openframe-data-nats`, which half the platform depends on. Adding a second provider later is a new implementation, not a broadcaster change.

**Push is a sink of its own, not a NATS fallback.** It fires even when NATS is disabled, because the two cover different states: sockets reach a foreground client, push reaches a backgrounded or killed one. Humans only — machines are agents, not phones.

**Auth is Application Default Credentials, not a service-account key.** The GCP org enforces `iam.disableServiceAccountKeyCreation`, so a key cannot even be issued. The same `getApplicationDefault()` call resolves Workload Identity in GKE and `gcloud` ADC locally — nothing to store, rotate, or leak, and no k8s secret. (The pod SA already had Workload Identity; it has been granted `roles/firebasecloudmessaging.admin` on the Firebase project.)

**The data payload carries the whole serialized context**, not a curated set of routing fields — the client owns deep-linking and can change it without a backend release. Oversized contexts are dropped rather than losing the push to FCM's ~4KB limit.

## Correctness details worth a look

- **Token re-association:** registering an existing token upserts it onto the caller, so a phone handed to another user after logout stops receiving the previous user's pushes.
- **Dead-token pruning is conservative:** only `UNREGISTERED` / `INVALID_ARGUMENT` / `SENDER_ID_MISMATCH` drop a row. A transient `UNAVAILABLE` or `QUOTA_EXCEEDED` must not cost a user their device registration.
- **Tenant safety:** nothing in the platform populates `tenantId` on insert, so the repository never inserts an entity directly — it upserts through `TenantAwareMongoTemplate`, letting MongoDB build the row from the query's equality conditions. Document and read-filter therefore cannot drift apart. Uniqueness is `{tenantId, token}` (the convention `AuthUser` and `MongoRegisteredClient` already use), which today degenerates to uniqueness on the token and stays correct if tenant-isolation ever shares collections.
- **Best-effort throughout:** a dead provider never fails the notification, the caller, or socket delivery.

## Tests

`openframe-data-nats` 53 green (incl. 4 new: push fires per admin never per machine; fires without NATS; absent bean is a no-op; a throwing sender is swallowed and the other recipients still get pushed). `openframe-notification-push` 7 green (no devices = no-op; multicast; UNREGISTERED pruned; UNAVAILABLE *not* pruned; whole-batch failure swallowed; payload contract; oversized context dropped). ArchUnit repository rules pass.

Wiring (registration endpoint + config) lands in openframe-saas-tenant on top of this.

## Summary by CodeRabbit

* **New Features**
  * Added FCM push support for iOS and Android, including device token registration and dead-token cleanup.
  * Introduced per-user push settings (default enabled) to allow disabling push recipients.
  * Added best-effort out-of-band channel delivery for admin recipients, with web-grace logic to avoid re-sending when already read.
  * Added configurable FCM payload budgeting (title/body/context) and delivery timeouts with startup validation.
* **Bug Fixes**
  * Channel delivery is resilient: failures no longer interrupt in-app notification flow.
  * Disabled push recipients are excluded from channel delivery.


openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationBroadcaster.java
kirill-567, line 79: isn't it chanel?

openframe-data-mongo-common/src/main/java/com/openframe/data/document/push/PushDevice.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.data.annotation.Id;
+import org.springframework.data.mongodb.core.index.CompoundIndex;
+import org.springframework.data.mongodb.core.index.CompoundIndexes;
+import org.springframework.data.mongodb.core.mapping.Document;
+
+import java.time.Instant;
+
+@Document(collection = "push_devices")
+@CompoundIndexes({
+        @CompoundIndex(name = "tenant_token_unique", def = "{'tenantId': 1, 'token': 1}", unique = true),
```
kirill-567, line 17: What's the idea of this index?

openframe-data-mongo-common/src/main/java/com/openframe/data/document/push/PushDevice.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private String tenantId;
+
+    private String userId;
+
+    private String token;
+
+    private PushPlatform platform;
+
+    private Instant createdAt;
```
kirill-567, line 37: Use mongo annotations.

openframe-data-mongo-sync/src/main/java/com/openframe/data/repository/push/impl/CustomPushDeviceRepositoryImpl.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.data.mongodb.core.query.Update;
+
+import java.time.Instant;
+import java.util.Collection;
+import java.util.List;
+
+/** Writes use {@code upsert}, not {@code save}: only an upsert copies {@code tenantId} from the filter onto the new row. */
+@Slf4j
+@ConditionalOnProperty(name = "openframe.tenant-isolation.enabled", havingValue = "true")
+public class CustomPushDeviceRepositoryImpl extends TenantAwareRepositorySupport implements CustomPushDeviceRepository {
```
kirill-567, line 25: Why custom repository instead of default Spring Data interface.

openframe-notification-push/src/main/java/com/openframe/notification/push/FcmProperties.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.time.Duration;
+
+@Data
+@ConfigurationProperties(prefix = "openframe.push.fcm")
+public class FcmProperties implements InitializingBean {
+
+    static final int FCM_PAYLOAD_LIMIT_BYTES = 4096;
```
kirill-567: configs

openframe-notification-push/src/main/java/com/openframe/notification/push/FcmPushSender.java
```diff
@@ -0,0 +1,170 @@
...
+    private final ObjectMapper objectMapper;
+    private final FcmProperties properties;
+
+    @Override
+    public String name() {
+        return "fcm";
+    }
+
+    @Override
+    public void deliver(String userId, Notification notification, NotificationCategory category) {
```
kirill-567, line 63: Can you describe how this logic works and why we have tokens limit ?

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationBroadcaster.java
semen-flamingo: NATS deliberately isn't a NotificationChannel: it serves machines too (channels are human-only — a machine has no phone/Slack), and it carries a richer contract (publishUpdateToUser/Machine for live card upserts on update()) that FCM/Slack can't honor. Folding it into the channel port would force every future channel to implement machine delivery and update semantics it can't support. Failure semantics differ too: a missed socket publish is reconciled via GraphQL catch-up; a missed push is just dropped.

openframe-data-mongo-common/src/main/java/com/openframe/data/document/push/PushDevice.java
semen-flamingo: Two jobs. (1) Uniqueness: one row per token per tenant — registerToken upserts by token, and the unique index is what makes a concurrent double-registration collapse into a single row instead of duplicates (the DuplicateKeyException handling in the impl exists exactly for losing that insert race). It also makes re-binding a token to a new user (phone handed over / re-login) an atomic overwrite of one row. (2) The tenant prefix matches how every query is scoped: the collection is shared, and the tenant-aware template adds tenantId to each filter, so lookups and deletes by token ride this index.

openframe-notification-push/src/main/java/com/openframe/notification/push/FcmPushSender.java
semen-flamingo: Flow: load the user's device rows → chunk tokens into groups of ≤500 → one sendEachForMulticast per chunk → collect per-token failures whose error code is UNREGISTERED/SENDER_ID_MISMATCH (token permanently dead) → delete those rows so we stop sending to them. 

The 500 is FCM's own hard cap on a MulticastMessage: build() throws IllegalArgumentException above it, so without chunking a user with 501+ device rows would get zero pushes (the build fails before anything is sent). 

INVALID_ARGUMENT is deliberately not treated as a dead token — FCM returns it for an oversized payload too, which fails for every token at once and would wipe all of a user's devices over one bad message.

openframe-data-mongo-sync/src/main/java/com/openframe/data/repository/push/impl/CustomPushDeviceRepositoryImpl.java
semen-flamingo: We need some custom logic in the repo which Spring Data cant handle. But I tried to use it as much as I can (e.g. findByUserId)

openframe-notification-push/src/main/java/com/openframe/notification/push/FcmProperties.java
semen-flamingo: Moved to constants

openframe-data-mongo-common/src/main/java/com/openframe/data/document/push/PushDevice.java
semen-flamingo: I didnt use it because of our custom repo logic. I made a research with claude and here is what he said:

Auditing is enabled platform-wide (MongoInfraConfig's @EnableMongoAuditing), but it fires via AuditingEntityCallback — a BeforeConvertCallback — and MongoTemplate invokes entity callbacks only on entity writes (doInsert/doSave/insert-batch/findAndReplace). The Update path (updateFirst/updateMulti/upsert → doUpdate) never touches callbacks or auditing — verified against spring-data-mongodb 4.3.0 sources; the only entity metadata it honors is @Version. It's a known, still-open enhancement since 2014: spring-projects/spring-data-mongodb#1797 (#2399 closed as duplicate). Every write to this collection is an Update-based upsert (atomic insert-or-rebind on the unique index), so @CreatedDate/@LastModifiedDate would compile and never populate — manual $setOnInsert createdAt / $set updatedAt is the only variant that works on this path.

openframe-data-mongo-sync/src/main/java/com/openframe/data/repository/push/impl/CustomPushDeviceRepositoryImpl.java
kirill-567: I see no methods that can not be implemented using interface

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
```diff
@@ -0,0 +1,135 @@
...
+ * A separate bean, not a method on NotificationBroadcaster: @Async only fires through a Spring proxy,
+ * so a self-invoked @Async method would run inline on the caller's thread — the very thing this moves
+ * off it. A channel does blocking network I/O; the notification is already persisted and on the socket.
+ */
+@Service
+@Slf4j
+public class NotificationChannelDispatcher {
+
+    private final List<NotificationChannel> channels;
+    private final Optional<NotificationSettingsRepository> settingsRepository;
```
kirill-567, line 34: Why optinal?

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
```diff
@@ -0,0 +1,135 @@
...
+                recipients.forEach(userId -> deliverSafely(channel, userId, notification, category)));
+    }
+
+    /** One parked virtual thread per notification — cheap. False only on interrupt (shutdown): drop, don't half-send. */
+    private boolean awaitGrace() {
+        if (webGraceSeconds <= 0) {
+            return true;
+        }
+        try {
+            Thread.sleep(Duration.ofSeconds(webGraceSeconds));
```
kirill-567, line 74: What's this logic?

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
semen-flamingo: The NotificationSettingsRepository bean only exists on tenant pods (it's @TenantAwareRepository, gated by openframe.tenant-isolation.enabled). openframe-data-nats is consumed by services that don't enable tenant isolation, so a hard dependency would fail their context startup

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
semen-flamingo: This is logic that we discussed with Oleksandr. This one makes a delay for push notification (15s default) and if it is being read (user is currently active in browser and we dont need to spam on his phone)

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
kirill-567: That's on prod for a long ago.

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
kirill-567: Thread.sleep with constant value looks like hack for me.
Please create tech dept ticket. Should be found something better.

kirill-567 (approved):

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
semen-flamingo: The `NotificationSettingsRepository` bean only exists on tenant pods (it's `@TenantAwareRepository`, gated by `openframe.tenant-isolation.enabled`). `openframe-data-nats` is consumed by services that don't enable tenant isolation, so a hard dependency would fail their context startup. `Optional` lets them boot with settings-based muting simply absent — treated the same as "no document": everyone enabled. Same reason the NATS publisher is `Optional` here.

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationChannelDispatcher.java
semen-flamingo: It's a stopgap to avoid double-notifying a user who's active in the browser, until real presence lands. On broadcast the socket reaches a foreground client instantly, but we also fire a phone push — annoying if they're already looking at it. So the push waits `openframe.push.web-grace-seconds` (default 15, 0 = off), then re-reads the read-state rows and drops recipients who read/dismissed it on the web within the window (an approval resolved by someone else flips everyone's row, so that self-suppresses). Fail-open: a missing row or a failed lookup still gets the push. It runs on the async virtual thread, so it never blocks the caller. FCM has no server-side delay+cancel, so it's built here — same shape Slack/Discord use.

openframe-data-mongo-sync/src/main/java/com/openframe/data/repository/push/impl/CustomPushDeviceRepositoryImpl.java
semen-flamingo: You're right — pushed. Only `registerToken` stays custom now: an upsert with insert-or-rebind race handling on the `{tenantId, token}` unique index, which isn't expressible as a derived query. The token deletes moved to derived `deleteByUserIdAndToken` / `deleteByTokenIn`, and `findByUserId` is already derived. On the scoping worry that kept me from doing this: derived deletes returning a count ARE tenant-scoped — `DeleteExecution` routes through `operations.remove(Query, Class, String)`, which `TenantAwareMongoTemplate` overrides and scopes (unlike derived finds, no `query()` proxy needed). Verified against spring-data-mongodb 4.3.2 sources and exercised live against real Mongo with tenant-isolation on.

openframe-notification-push/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+        <connection>scm:git:https://github.com/flamingo-stack/openframe-oss-lib.git</connection>
+        <developerConnection>scm:git:ssh://git@github.com/flamingo-stack/openframe-oss-lib.git</developerConnection>
+        <tag>HEAD</tag>
+    </scm>
+
+    <properties>
+        <firebase-admin.version>9.9.0</firebase-admin.version>
+        <!-- Must match what firebase-admin's google-auth-library-oauth2-http resolves to. -->
+        <google-auth.version>1.43.0</google-auth.version>
+    </properties>
```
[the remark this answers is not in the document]
semen-flamingo: The pin isn't a CVE remediation — it fixes a classpath split (firebase-admin's `oauth2-http:1.43.0` was resolving against `credentials:1.23.0` pulled at a shallower depth, causing `NoClassDefFoundError` on the first token refresh). No CVE is named against either artifact; transitive-CVE hygiene is tracked repo-wide by dependabot.

PR #1486, opened 2026-07-17
fix(users): return the same role set from /me and /users/{id}

kanashenko:
https://app.clickup.com/t/9013925967/86ajh7h2p

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated role handling so users with the Owner role also receive the Admin role where applicable.
  * Ensured consistent effective roles across user responses and authorization tokens.
  * Improved handling when no roles are assigned.

### DIFF
--- a/openframe-data-mongo-common/src/main/java/com/openframe/data/document/user/UserRole.java
+++ b/openframe-data-mongo-common/src/main/java/com/openframe/data/document/user/UserRole.java
@@ -1,6 +1,26 @@
 package com.openframe.data.document.user;
 
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
 public enum UserRole {
     ADMIN,
-    OWNER
+    OWNER;
+
+    /**
+     * Adds ADMIN when the user is an OWNER.
+     *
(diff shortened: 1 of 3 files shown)


openframe-data-mongo-common/src/main/java/com/openframe/data/document/user/UserRole.java
```diff
@@ -1,6 +1,26 @@
...
-}
\ No newline at end of file
+    OWNER;
+
+    /**
+     * Adds ADMIN when the user is an OWNER.
+     *
+     * <p>Mongo stores only what was granted, so an owner is saved as just OWNER.
+     */
+    public static Set<UserRole> effective(Collection<UserRole> granted) {
```
aliaska-varieva, line 16: what if just like
```public static Set<UserRole> effective(Collection<UserRole> granted) {
    var effective = granted == null
            ? new LinkedHashSet<UserRole>()
            : new LinkedHashSet<>(granted);
    if (effective.contains(OWNER)) {
        effective.add(ADMIN);
    }
    return effective;
}

aliaska-varieva (approved):

PR #1581, opened 2026-07-28
feat(chat): server-issued conversation ids + server-side history hydration in the SSE adapter

michaelassraf:
## What changed

The SSE/Guide chat adapter's memory model is inverted to match the hub's new server-side conversation store (pairs with flamingo-stack/multi-platform-hub#958):

- **No client-side message store.** localStorage keeps ONLY `{ conversationId }` under an unversioned key (`mingo-chat-<source>[-u-<email>]`); the full-history persistence (`PersistedChatState` with messages/sources/refs/sendCount) is deleted, and old keys are swept with a boundary-aware rule (`flamingo` can never sweep `flamingo-teaser` keys).
- **Ids are server-issued, never client-generated.** The first send of a session goes out without an id; the adapter captures `metadata.conversationId` from the leading metadata frame, persists it, and echoes it on every later turn. `generateThreadId` is deleted. "New chat" (`clearMessages`) just drops the stored id — the next send mints a fresh conversation server-side.
- **The wire carries only the newest user message.** All history-replay plumbing is removed (`apiMessages` assembly, hidden-message filtering, `messagesRef` threading, `flattenAssistantContent` usage in the stream fn). The approval branch sends `{ proposal_id, action, conversationId }` — no messages.
- **Mount-time hydration** rebuilds messages/inline-card refs/send counter from `GET <chatStreamUrl>/history?conversationId=…` (new optional `chatHistoryUrl` runtime endpoint, defaulting to `<chatStreamUrl>/history` so reverse-proxy embedders work unchanged). New `hydrateMessages` primitive in `useChat`; `UnifiedChatState` gains optional `isHydratingHistory`; `ChatTurnMeta` gains `conversationId`.

## Reviewer notes

- Breaking pairing: needs the hub side (multi-platform-hub#958) deployed for persistence to function; against an older hub the chat still works but every turn is single-message context.
- `hydrateMessages` prepends server history if the user typed before hydration landed — nothing typed is lost, and correctness is unaffected either way because the server resolves history from its own store.
- Verified with standalone tsc checks on the touched files; full type-check/build runs in this repo's CI.

## Summary by CodeRabbit

- **New Features**
  - Chat conversations now restore server-saved history when reopened.
  - Conversation history is linked by conversation ID for more reliable persistence.
  - Added a loading state while chat history is being restored.
  - Chat hosts can configure a dedicated history endpoint.

- **Bug Fixes**
  - Restored assistant references and message context now align correctly with previous conversation turns.
  - Clearing a chat also resets its saved conversation state.

### DIFF
--- a/openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
+++ b/openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
@@ -40,16 +40,21 @@
  */
 
 import { useState, useEffect, useCallback, useMemo, useRef, type MutableRefObject } from 'react'
-import { useChat, type Message, type StreamFnExtraOptions } from './use-chat'
+import { useChat, type StreamFnExtraOptions } from './use-chat'
 import { useRequiredChatRuntime } from '../../../contexts/chat-runtime-context'
 import type { ChatRef } from '../chat-ref.types'
 import { buildChatRefKey } from '../types/chat.types'
 import type { MessageSegment } from '../types/message.types'
 import { useSlashCommandRegistry, type SlashCommandSummary } from './use-slash-commands'
-import { getChatProxyAuth } from '../utils/chat-proxy-auth-storage'
 import { chatAuthedFetch } from '../utils/chat-authed-fetch'
+import {
+  createChatConversationStorage,
+  pruneStaleChatConversationStorage,
+  type PersistedChatConversation,
+} from '../utils/chat-conversation-storage'
+import type { LocalStorageAdapter } from '../../../utils/local-storage-adapter'
+import { useChatHistoryHydration } from './use-chat-history-hydration'
 import { parseScrollAnchor, type ScrollAnchor } from '../utils/scroll-anchor'
-import { AUTO_CONTINUATION_DIRECTIVE_PREFIX } from '../utils/auto-continuation-directive'
 import { sanitizeTitleForChat } from '../utils/slash-dispatch-utils'
 import { defaultTableIdForDocumentType } from '../utils/source-icons'
 import type {
@@ -226,6 +231,7 @@ function createDocStreamFn(
   bumpMetaTick: () => void,
   sendCountRef: MutableRefObject<number>,
   conversationIdRef: MutableRefObject<string | null>,
+  conversationStorage: LocalStorageAdapter<PersistedChatConversation>,
 ) {
   // CRITICAL: the decoder + buffer MUST live INSIDE the returned async-
   // generator function (per-call closure), NOT at the factory level. A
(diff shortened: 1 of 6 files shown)


openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
```diff
@@ -633,49 +646,38 @@ function pruneStaleChatStorage(source: DocSource): void {
...
   if (typeof window === 'undefined') return
   try {
-    window.localStorage.setItem(chatStorageKey(source), JSON.stringify(state))
+    window.localStorage.setItem(chatStorageKey(source), JSON.stringify({ conversationId }))
   } catch {
     // Quota exceeded or private mode — silently drop.
   }
 }
 
+function clearPersistedConversationId(source: DocSource): void {
```
michaelassraf: should use our normal local storage manager

openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
```diff
@@ -792,37 +794,104 @@ export function useSseChatAdapter(
...
+    const conversationId = conversationIdRef.current
+    if (!conversationId) return
+    const key = `${source}:${conversationId}`
+    if (hydratedKeyRef.current === key) return
+    hydratedKeyRef.current = key
+    let cancelled = false
+    setIsHydratingHistory(true)
+    ;(async () => {
+      try {
+        const res = await chatAuthedFetch(
```
michaelassraf: should move to proper hook class

openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
michaelassraf: Fixed in ed28101f — conversation-id persistence now lives in `components/chat/utils/chat-conversation-storage.ts` built on the lib-standard `createLocalStorageAdapter` (SSR guard, try/catch, quota logging, dynamic namespace so a proxy-auth identity switch re-partitions the key automatically). The raw `window.localStorage` helpers are deleted; only the stale-key sweep still enumerates raw storage, since Web Storage exposes iteration only on the raw object.

openframe-frontend-core/src/components/chat/hooks/use-sse-chat-adapter.ts
michaelassraf: Fixed in ed28101f — the inline hydration effect is extracted to a dedicated hook, `components/chat/hooks/use-chat-history-hydration.ts` (`useChatHistoryHydration`), owning the fetch, row→Message mapping, refs/send-counter rebuild, and the once-per-conversation guard. The adapter now just calls the hook and wires its `isHydratingHistory` / `hydratedKeyRef` outputs.

PR #1618, opened 2026-07-31
Artifact validation gate: validators, script-service gate, validation metadata

andriikachur:
## Summary

Part A of the library validation gate (bug: Mingo saves unvalidated queries, policies, and scripts to the shared library):

- **Validation engine** (`com.openframe.api.service.validation.artifact`): `OsquerySqlValidator` (sqlite compile, single SELECT/WITH only), `ScriptSyntaxValidator` (parse-only `bash -n` / `sh -n` / python `ast.parse`, honest `SYNTAX_SKIPPED_NO_INTERPRETER` for shells without a parser in the container), `StaticSafetyAnalyzer` (catastrophic patterns block; reboot/service-stop/HKLM/hardcoded credentials flag high-impact)
- **Mandatory gate in `ScriptService.create/update`** — the single choke point shared by the GraphQL UI mutation and the AI agent; no code path saves a script without validation. `update` now takes `updatedBy` to record approvals.
- **High-impact artifacts require a recorded human approver** — the `AI_AGENT` sentinel cannot self-approve.
- **Validation metadata**: `Script.validation` embedded doc, `artifact_validations` collection (for Fleet-side artifacts, written by the AI agent in the saas-tenant part), GraphQL `Script.validation` type.

New dependency: `org.xerial:sqlite-jdbc` (osquery speaks the SQLite dialect).

Part B (Fleet runners gate + osquery live-query dry-run + guardrails-on-save) lands in openframe-saas-tenant and depends on this PR plus a one-line saas-lib change.

## Tests
80 tests green across affected modules, incl. 30 new validator/gate tests; existing ScriptService tests now run through the real gate.

## Summary by CodeRabbit

* **New Features**
  * Added automatic validation for scripts and osquery SQL during creation and updates.
  * Detects syntax errors, destructive actions, credential exposure, and other safety concerns.
  * Blocks invalid artifacts and requires human approval for high-impact scripts.
  * Exposes validation status, warnings, target platforms, methods, and approval details.
  * Preserves compatibility with scripts created before validation was introduced.

* **Bug Fixes**
  * Updates now associate validation results with the authenticated actor.

* **Tests**
  * Added comprehensive coverage for syntax, safety, approval, feature configuration, and SQL validation scenarios.


openframe-api-lib/src/main/java/com/openframe/api/dto/rmm/script/ScriptValidationResponse.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+/**
+ * API representation of the validation gate outcome stored on a script.
+ * Null on the parent {@link ScriptResponse} only for scripts created before
+ * the validation gate shipped.
+ */
+@Data
+@Builder
+public class ScriptValidationResponse {
```
kirill-567: Why api lib?

kirill-567: Seams it also should be saas because it's ai related feature.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>Delegates to the shell's own parser where one exists in the container
+ * ({@code bash -n} / {@code sh -n} / python {@code ast.parse}); nothing is
+ * ever executed. Shells with no interpreter available in the JVM container
+ * (PowerShell, CMD, NuShell — and python when python3 is absent) are recorded
+ * as {@code SYNTAX_SKIPPED_NO_INTERPRETER} rather than blocked: the static
+ * safety rules still run, and the honest skip shows up in artifact metadata.
+ */
+@Slf4j
+@Component
+public class ScriptSyntaxValidator {
```
kirill-567: Static analisation doesn't work. Low chance to catch anything.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
```diff
@@ -0,0 +1,108 @@
...
+            case PYTHON -> parseWithStdin("SYNTAX_PYTHON_AST", body,
+                    "python3", "-c", "import sys, ast; ast.parse(sys.stdin.read())");
+            case POWERSHELL, CMD, NUSHELL -> skipped();
+        };
+    }
+
+    private ArtifactValidationResult parseWithFile(String method, String body, String... command) {
+        Path tmp = null;
+        try {
+            tmp = Files.createTempFile("of-script-validate", ".tmp");
```
kirill-567: Why do we have file creation. 
Generally bad practice for stateless service.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private ArtifactValidationResult parseWithStdin(String method, String body, String... command) {
+        return runParser(method, new ProcessBuilder(command), body);
+    }
+
+    private ArtifactValidationResult runParser(String method, ProcessBuilder pb, String stdin) {
+        pb.redirectErrorStream(true);
+        try {
+            Process p = pb.start();
```
kirill-567: This is big hole for the attack.
I expected dry run on small amount of devices.

openframe-api-lib/src/main/java/com/openframe/api/dto/rmm/script/ScriptValidationResponse.java
andriikachur: The gate isn't AI-specific, which is why it sits here rather than in saas-lib. It lives inside `ScriptService`, and that service backs **both** the AI agent tools and the `createScript`/`updateScript` GraphQL mutations the technician UI calls — that is the whole point of the placement: there is no save path around it. The scripts library is also an OSS feature (`openframe-api` in openframe-oss-tenant runs the same `openframe-api-service-core`), so moving this to saas would leave the UI path and the OSS install unguarded, or force the logic to exist twice.

The parts that genuinely are AI-specific — the Fleet policy/query runners, the save-time guardrails, the prompt rules — all live in openframe-saas-tenant.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
andriikachur: Agreed — and after your feedback it is no longer the mechanism we rely on.

The design changed: the server executes nothing, and the service images ship **without** bash/python on purpose, so in production this validator always records `SYNTAX_SKIPPED_NO_INTERPRETER`. An artifact is proven by actually running it on machines the technician approves (`runCommandV2` for scripts, `runLiveQuery` for osquery), each under its own approval card. What still blocks a save server-side is the osquery SQL parse (sqlite-jdbc, in-process, no subprocess) and the static rules for catastrophic patterns.

The class is kept as an opt-in for anyone who deliberately adds an interpreter (a pwsh sidecar was the case we had in mind), and the metadata is explicit when it does not run. If you'd rather it were deleted outright, I have no objection — say the word and it goes.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
andriikachur: Fair point — fixed in 34913a07. `bash -n` / `sh -n` read the script from stdin just as well, so the body is piped in and no file is written at all. That also removed a whole failure mode: temp-file creation could fail and was being reported as "interpreter unavailable", which was wrong on its own.

openframe-api-lib/src/main/java/com/openframe/api/service/validation/artifact/ScriptSyntaxValidator.java
andriikachur: That is exactly where it ended up, and your comment is a large part of why.

Server-initiated execution is gone — `OsqueryDryRunService` was deleted along with the whole idea of the backend running anything on customer machines. What remains server-side never executes: an in-process sqlite parse of osquery SQL and regex-based static rules.

Verification now happens on a small, explicit set of devices, as you expected — but chosen and approved by the technician rather than by the backend: the agent runs the exact artifact through `runLiveQuery` / `runCommandV2` on machines the technician names, and each run raises its own approval card before it executes.

Verified end-to-end on a PR tenant with one enrolled machine: two approval cards (one for the test run, one for the save), the saved body byte-identical to the tested one, and a refusal to save after a failed run.

PR #1664, opened 2026-08-06
feat(notification): read-state transitions emit events to an SPI listener list

semen-flamingo:
Read/dismiss/delete transitions used to change Mongo and tell no one — cross-device push retraction and live web read-sync (ClickUp 86ajrmwkf) have nothing to hook into.

This adds the seam, same pattern as the `NotificationChannel` SPI:

- **`NotificationReadEventListener`** — SPI called after every PERSISTED transition; `List` injection, so the OSS edition (no impls) is a zero-iteration no-op.
- **`NotificationReadEvent`** — recipientId, recipientType, affected notificationIds, transition (`READ` | `DELETED`).
- Events fire only on **actual** flips: a repeated markRead matches nothing and stays silent — retraction must not re-trigger on idempotent repeats.
- Bulk ops (`markAllAsRead`, `deleteAllRead`) snapshot the affected ids **before** the flip and emit ONE event — under concurrency the snapshot can over-/under-report (deliberate: events are best-effort, a throwing listener already loses one, so listeners must be idempotent; exactness would cost a transaction); `dismissForAllRecipients` (lifecycle resolve) emits per previously-UNREAD recipient.
- Listeners are synchronous and **fail-open**: a throwing listener is logged and skipped, never affecting the mutation result or the other listeners.

No behaviour change for the OSS edition. The SaaS retraction listener (outbox cancel / tombstone pushes) and the NATS live-update publisher come as separate downstream PRs.

https://claude.ai/code/session_01BKdcCJmnHipDNZe1qYqp9H

## Summary by CodeRabbit

* **New Features**
  * Added notification read-state events for read and deleted transitions.
  * Added support for reacting to individual and bulk notification read-state changes.
  * Added recipient-based filtering of notification read-state records.

* **Bug Fixes**
  * Listener failures no longer interrupt notification read-state updates.
  * Bulk operations now report all affected notifications accurately.


openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/.NotificationReadStateService.md
```diff
@@ -1,5 +1,5 @@
 <!-- source-hash: 8e455be47104c50ad5fe974c09cd9cfe -->
```
kirill-567, line 1: We don't touch this files as it's auto generated documentation.

openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/NotificationReadStateService.java
```diff
@@ -61,11 +62,21 @@ public List<NotificationReadState> findRecipients(@NotBlank String notificationI
...
 
     public long markAllAsRead(@NotBlank String recipientId, @NotNull RecipientType recipientType) {
-        return repository.markAllAsRead(recipientId, recipientType);
+        // Snapshot BEFORE the bulk flip: a row created in between gets flipped but not reported —
+        // acceptable, its banner is cleaned up by the client's on-open reconciliation.
+        List<String> unreadIds = notificationIds(
+                repository.findByRecipientIdAndRecipientTypeAndStatus(recipientId, recipientType, ReadStatus.UNREAD));
+        long flipped = repository.markAllAsRead(recipientId, recipientType);
+        publish(recipientId, recipientType, unreadIds, NotificationReadEvent.Transition.READ);
+        return flipped;
```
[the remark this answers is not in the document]
kirill-567: valid?

openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/NotificationReadStateService.java
```diff
@@ -61,11 +62,21 @@ public List<NotificationReadState> findRecipients(@NotBlank String notificationI
     public boolean markRead(@NotBlank String recipientId,
                             @NotNull RecipientType recipientType,
                             @NotBlank String notificationId) {
-        return repository.markAsRead(recipientId, recipientType, notificationId) > 0;
+        boolean transitioned = repository.markAsRead(recipientId, recipientType, notificationId) > 0;
+        if (transitioned) {
+            publish(recipientId, recipientType, List.of(notificationId), NotificationReadEvent.Transition.READ);
```
kirill-567, line 67: I would prefer to avoid spring events. 
If there reason to do it ?

openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/.NotificationReadStateService.md
semen-flamingo: Fixed

openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/NotificationReadStateService.java
semen-flamingo: Thats not spring event

openframe-data-mongo-sync/src/main/java/com/openframe/data/service/notification/NotificationReadStateService.java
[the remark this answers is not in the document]
semen-flamingo: Even though this is technically valid
This race is harmless for us currently as all consumers are ok with it. We already supported at least once delivery and this race can only produce duplicates which are ok in this implementation

kirill-567 (approved):

PR #1689, opened 2026-08-10
feat(notification): settings v2 — one master switch over every channel, per-group checkboxes

semen-flamingo:
Settings v2 for the notifications modal (ClickUp 86ajvumfg): the master switch now covers **all** notification delivery — Semen's call: no separate push toggle; browser-level notifications live in the FE sidebar, phone-level in OS app settings.

**Where the toggles bite — the audience, not the channels.** `NotificationBroadcaster` drops opted-out admins before `createForAudience`: no read-state row → no card, no badge, no NATS message, no push, and nothing arrives retroactively after re-enabling. The dispatcher's `withoutMuted` is deleted — audiences arrive pre-filtered. Machines are untouched (settings are a human thing).

**Groups = the modal's checkboxes** (`NotificationSettingGroup`: TICKET_ASSIGNED, TICKET_STATUS_CHANGED, CUSTOMER_REPLIED, ADMIN_REPLIED, MINGO_MESSAGES, APPROVAL_TICKET, APPROVAL_MINGO). Which group a notification belongs to is the **descriptor's** call — `settingsGroup(context)`, the same seam as `category(context)`; the SaaS descriptors get their mappings in a follow-up saas-lib PR (approval splits by nullable `ticketId`, exactly like its category). Default = null = no checkbox = cannot be muted — **deliberately fail-open**: a type whose descriptor forgot the mapping loses mutability, never delivery.

**Absence always means deliver**: no document, no `enabled` field, no `mutedGroups` entry (including groups added later) — all default to on. Zero migrations. The one exception is a **failed** settings lookup: the dispatch drops its admin audience (log.error) — consistent with the rest of broadcast, where a Mongo failure aborts delivery rather than delivering against unknown preferences.

**Contract v2 — clean break, no legacy** (Semen's call: the only `pushEnabled` consumer was the mobile-shell settings modal, and the app is not in production): `notificationSettings { enabled, typeSettings { group, enabled } }`; `updateNotificationSettings(enabled: Boolean!, typeSettings)` — omitting `typeSettings` keeps the stored mutes. Storage is `enabled` + a `mutedGroups` set — only muted groups persist (a map of constant falses is a set in disguise); the full per-group list is materialized at the API layer. The Mongo field, the schema member and the mutation argument are all gone; an absent `enabled` simply means on. ⚠️ The mobile-shell modal breaks against this schema until it migrates to v2 — part of the FE settings ticket.

Until the saas descriptor PR lands, every type is ungrouped → behavior is unchanged except the master switch (which now means what the modal says it means).

https://claude.ai/code/session_01BKdcCJmnHipDNZe1qYqp9H

## Summary by CodeRabbit

* **New Features**
  * Added a master notification switch and configurable preferences for individual notification categories.
  * Notification settings can now be viewed and updated together, including category-specific choices.
  * Notifications respect administrator preferences before delivery, while machine notifications remain unaffected.

* **Bug Fixes**
  * Notification delivery defaults to all eligible recipients when preferences are unavailable or cannot be resolved.
  * Unmapped notification types continue to be delivered.


openframe-data-mongo-common/src/main/java/com/openframe/data/document/notification/NotificationSettings.java
```diff
@@ -27,10 +28,33 @@ public class NotificationSettings implements TenantScoped {
...
     private Instant createdAt;
 
     private Instant updatedAt;
+
+    public boolean masterEnabled() {
+        return enabled != null ? enabled : pushEnabled;
+    }
+
+    public boolean groupEnabled(NotificationSettingGroup group) {
+        return group == null || typeSettings == null || typeSettings.getOrDefault(group, true);
```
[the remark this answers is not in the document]
semen-flamingo: Obsolete since 2a202b28: storage moved from Map<group, Boolean> to Set<NotificationSettingGroup> mutedGroups — there is no unboxing left; the policy checks membership (!mutedGroups.contains(group)).

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationBroadcaster.java
```diff
@@ -51,7 +57,7 @@ public Notification broadcast(NotificationCommand command) {
         log.debug("Persisted notification {} (admins={}, machines={})",
                 saved.getId(), command.getAdminAudience().size(), command.getMachineAudience().size());
 
-        Set<String> admins = command.getAdminAudience();
+        Set<String> admins = withoutOptedOut(command.getAdminAudience(), command.getContext());
```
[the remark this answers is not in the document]
semen-flamingo: Fixed in c6ca4e32: audience filtering now runs before persistence — when both audiences end up empty the dispatch logs and returns without saving, so no invisible orphan docs. Covered by fully_muted_dispatch_persists_nothing.

openframe-data-nats/src/main/java/com/openframe/data/nats/service/NotificationBroadcaster.java
```diff
@@ -92,6 +98,36 @@ public Notification broadcast(NotificationCommand command) {
...
+                return admins;
+            }
+            Set<String> kept = new HashSet<>(admins);
+            for (NotificationSettings row : rows) {
+                if (!row.allows(group)) {
+                    kept.remove(row.getUserId());
+                }
+            }
+            return kept;
+        } catch (RuntimeException ex) {
```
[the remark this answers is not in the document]
semen-flamingo: Taken in c6ca4e32: settingsGroupOf now resolves inside the guarded block, so a throwing descriptor degrades into the standard drop-admins path instead of failing broadcast after persistence. Note the failure policy flipped to fail-closed earlier (ce4d043f, Semen's call): a settings failure drops the admin audience rather than delivering against unknown preferences. Covered by settings_group_resolution_failure_drops_all_admins.

openframe-data-mongo-common/src/main/java/com/openframe/data/document/notification/NotificationContextDescriptor.java
```diff
@@ -22,4 +22,9 @@ default NotificationCategory category() {
     default NotificationCategory category(NotificationContext context) {
         return category();
     }
+
+    /** Null = no checkbox = cannot be muted — a type whose descriptor forgot the mapping loses mutability, never delivery. */
+    default NotificationSettingGroup settingsGroup(NotificationContext context) {
+        return null;
```
kirill-567: What cases have null settings group ?

openframe-api-service-core/src/main/java/com/openframe/api/dto/NotificationSettingsView.java
```diff
@@ -0,0 +1,31 @@
...
+
+// Defaults are collapsed server-side: every group exactly once — the client never re-implements the defaulting rules.
+@Getter
+@AllArgsConstructor
+public class NotificationSettingsView {
+
+    private final boolean enabled;
+    private final List<NotificationTypeSetting> typeSettings;
+
+    public static NotificationSettingsView from(NotificationSettings settings) {
```
kirill-567: Mapper class or private method better than map model to dto at the model level.

openframe-api-service-core/src/main/resources/schema/notification-settings.graphqls
```diff
@@ -3,10 +3,34 @@ extend type Query {
...
 
 extend type Mutation {
-    updateNotificationSettings(pushEnabled: Boolean!): NotificationSettings!
+    "The modal saves full state; omitting typeSettings keeps the stored group overrides."
+    updateNotificationSettings(enabled: Boolean!, typeSettings: [NotificationTypeSettingInput!]): NotificationSettings!
+}
+
+"The 'Notify about' checkboxes. A notification type outside every group cannot be muted."
+enum NotificationSettingGroup {
+    TICKET_ASSIGNED
```
kirill-567, line 12: Who is responsible to convert it to text ?
FE? 
I prefer backend, but can you check best practices?

semen-flamingo: Checked common practice 
Usually its done on FE through i18n due to localsation and other concerns

kirill-567 (approved):

PR #1704, opened 2026-08-10
fix(api-lib): case-insensitive tag dedupe in TagService

semen-flamingo:
## Summary
Server-side half of [86ajcx9nh](https://app.clickup.com/t/86ajcx9nh), companion to flamingo-stack/openframe-saas-tenant#2478 (merged): tag keys differing only by letter case ("NEWTAG" vs "NewTag") are the same tag, but `TagService.createTag` deduped by exact `findByKeyAndEntityType` and the `tags` unique index (`tenantId+key+entityType`) has no collation — so a case-variant duplicate was physically creatable from any caller (Mingo `createTag` tool, `TagDataFetcher` UI mutation).

- `createTag` — dedupes case-insensitively: exact indexed lookup first, then an in-memory `equalsIgnoreCase` scan of the entity-type pool (pools are small; avoids regex-escaping pitfalls of a derived `IgnoreCase` Mongo query). Returns the existing tag instead of creating a near-duplicate.
- `updateTag` — rejects a rename that collides with a case-variant of *another* tag (`IllegalArgumentException`, which the Mingo tool runner already maps to a tool failure); changing the case of a tag's own key stays allowed.

## Testing
`TagServiceTest` (new): exact-match reuse, case-variant reuse without save, creation when no match, pool isolation across entity types, rename collision rejection, own-key case change allowed.

Note: existing case-duplicates already in a DB are not merged by this change — it only prevents new ones.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented creating or renaming tags with keys that duplicate existing keys, regardless of capitalization.
  * Allowed tags to change only in capitalization when they remain otherwise unique.
  * Ensured duplicate checks apply only to tags associated with the same entity type.


openframe-api-lib/src/main/java/com/openframe/api/service/TagService.java
```diff
@@ -244,6 +252,27 @@ public Tag updateTag(String tagId, String key, String description, String color)
         return tagRepository.save(tag);
     }
 
+    /**
```
kirill-567: No jdoc without reason please.

openframe-api-lib/src/main/java/com/openframe/api/service/TagService.java
```diff
@@ -208,7 +208,7 @@ public List<Tag> getTagsForMachine(String machineId) {
     public Tag createTag(String key, TagEntityType entityType, String description, String color) {
         log.info("Creating tag '{}' with entityType: {}", key, entityType);
 
-        Tag existing = tagRepository.findByKeyAndEntityType(key, entityType);
+        Tag existing = findByKeyIgnoreCase(key, entityType);
```
kirill-567, line 211: Can it be ignore of case at the spring data query ?

kirill-567 (approved):

openframe-api-lib/src/test/java/com/openframe/api/service/TagServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+        Tag result = tagService.createTag("NEWTAG", KNOWLEDGE_ARTICLE, null, null);
+
+        assertThat(result).isSameAs(newTag);
+        verify(tagRepository, never()).save(any());
+    }
+
+    @Test
+    void createTagSavesWhenNoCaseVariantExists() {
+        when(tagRepository.findByKeyAndEntityType("Printers", KNOWLEDGE_ARTICLE)).thenReturn(null);
+        when(tagRepository.findByKeyIgnoreCaseAndEntityType("Printers", KNOWLEDGE_ARTICLE)).thenReturn(List.of());
```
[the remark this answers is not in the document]
semen-flamingo: Pool isolation is enforced by the query itself now — both guards are `...AndEntityType` / `...AndEntityTypeAndIdNot` derived queries, so a case-variant in another pool is never even fetched; the mocks assert the exact entityType argument. Verified live on a feature tenant as well: with ticket tag `NewTag` present, `createTag(key: "NEWTAG", entityType: DEVICE)` created a new tag while the same key on `TICKET` returned 409.

Happy to add the explicit unit test as a small follow-up.

PR #1778, opened 2026-08-13
feat(chat): remove card registry modes — every type fetches by id

michaelassraf:
## What

The `CHAT_CARD_REGISTRY` `mode: 'fetch' | 'no-fetch'` split is deleted. One entry shape: `contentRefType` present = the card hydrates by id through `useChatCardItem` → `buildListUrl`, like every content card always did. Cards no longer depend on the SSE refs frame for display data — brokered streams and history replays self-heal.

- `refHydratedEntry` preset for the 16 ref-shaped types (github ×6, slack, tickets ×3, data-room, markdown, financials ×5): fetched ChatRef items overlay the marker's ref (`mergeFetchedRefOverlay` — fresh status beats a replayed ref's frozen one), `item.url` is the destination, `refFallback` renders from the server-built ref while the fetch is in-flight/erroring (no skeleton flash live, no vanish on transient failures; a confirmed fetch miss still tombstones).
- `buildListUrl` gains one builder per type via `ENTITY_CARD_ROUTES` — **one route per object** on the hub (`/api/github/commits`, `/api/slack-community/messages`, `/api/tickets/{internal,known-issues}`, `/api/docs/pages`, `/api/data-room/documents`, `/api/financials/*`; self tickets → `/api/tickets`). Internal/public github variants share their object's route. Byte-parity tests frozen.
- `faq` flips to fetch against its existing `/api/faqs?ids=` with a row-shape overlay; `extractItems` learns the `{ faqs }` response shape.
- Only `deleted_data` (tombstone) and `video` (body-embed hash ids, no backing row) stay ref-only — nothing exists server-side to fetch.
- New `MingoInfoCardSkeleton` for the refs-less loading state.

## Pairs with

flamingo-stack/multi-platform-hub#1080 — the per-object hydration routes + shared `makeEntityCardListHandler` factory.

## Tests

tsc clean; 556 chat/util tests green (incl. per-object byte-parity builders + shrunken null-type list).


openframe-frontend-core/src/components/chat/entity-cards/dispatch.tsx
```diff
@@ -345,6 +350,49 @@ function hubspotStatusToVariant(
   return 'grey'
 }
 
+/** Merge a fetched ChatRef-shaped item over the marker's ChatRef. The
```
michaelassraf: I dont wanna rely on chat ref at all, one single layer to maintain and it's api drive

michaelassraf: Done in c8396c37 — the ref-overlay layer and refFallback render paths are deleted. The fetched API item is now the single display-data layer (`fetchedItemDisplayRef` / `fetchedFaqDisplayRef`); the marker ref contributes only the loader-resolved url/targetPlatform for embed-origin routing. Loading = skeleton, same as every other fetch card.

openframe-frontend-core/src/utils/list-url.ts
```diff
@@ -76,6 +80,41 @@ const BUILDERS: Record<string, (ids: string[], base: string) => string> = {
...
+/**
+ * Per-OBJECT card-hydration routes for the types with no pre-existing
+ * public list API (one route per object, matching the repo convention).
+ * Each returns ChatRef-shaped items via the same source-binding +
+ * row-filter chain passive retrieval uses (a private object is simply
+ * not bound on public platforms → 404). The internal/public GitHub
+ * variants share ONE route per object — the route serves whichever
+ * variant is bound to the calling deployment's source.
+ */
+const ENTITY_CARD_ROUTES: Record<string, string> = {
```
michaelassraf: didnt we have this already somewhere? why do we have this hardcoded here? shouldnt it already exist?

michaelassraf: Fair — it was a second map sitting next to the map that already owns this job. Fixed in 4792ac4a: the 16 per-object URLs are now plain `BUILDERS` entries in the same pre-existing SSOT every other type lives in. And yes, this is the ONLY place these shapes exist — the hub's mappers/configs delegate here via `buildListUrl('<type>', ids)` (byte-parity tested), nothing else defines them.

PR #1783, opened 2026-08-13
Openframe Client Updater

danylo-babenko-flamingo:
# Port openframe-client-updater from oss-tenant

Ports the complete client-updater v2 work from `openframe-oss-tenant` (branch `hotfix/create-client-updater-service-v2`, PR #2180) into oss-lib as the new home of client development. The updater crate arrives verbatim from the oss-tenant branch tip (rustfmt-applied to satisfy this repo's `cargo fmt --check` CI), and the matching client-side changes are replayed onto `openframe-agent-lib`.

### New crate: `clients/openframe-client-updater`

A standalone OS service (`com.openframe.client-updater`) that owns the openframe-client binary swap. A second process is required because Windows locks a running executable — the client cannot replace itself. The updater owns the `CLIENT_UPDATE` JetStream consumer end to end; the client no longer participates in its own update.

**Hybrid token handling.** The updater first reuses the client's token (`shared_token.enc`, AES-GCM); if the client is down or the token is unusable, it falls back to its own OAuth2 `client_credentials` refresh using the shared `agent_config.json` credentials. Tokens live in memory only.

**Last-known-good ratchet with health-gated promotion.** Each update walks `Downloading → Verifying → StoppingService → ReplacingBinary → StartingService → VerifyingBoot → Observing`:
- the new binary must write a boot marker within 90s whose version matches the target (semver-normalized);
- a 10-minute observation window follows — a crash or failed health check inside it triggers automatic rollback;
- only after surviving observation is the version promoted to the last-known-good anchor, with a `.lkg` reserve copy of the binary kept next to the executable.

A failed update can therefore only fall back to the last *verified* version, never an arbitrarily old backup.

**Rollback support.** An incoming message with `rollback: true` bypasses the downgrade guard and, when the target version matches the reserve, restores it offline — no download needed.

**Crash recovery.** On startup the updater inspects persisted state (`updater_state.json`) and finishes whatever a mid-update crash interrupted — completing the swap, verifying boot, or rolling back — before connecting to NATS; the outcome is reported once connected.

### Client-side (`openframe-agent-lib`)

**In-client update machinery removed** (−3,400 lines): the old update service, state machine, PowerShell/shell update scripts, and updater launcher are gone. The client keeps exactly one update-related duty: installing and updating the updater itself as a tool.

**Update ordering.** During a client swap, all tool operations (install / uninstall / tool update / restart) are parked, not dropped: `park_or_dispatch` probes the updater's state file and, while a swap is in flight, holds each message with `AckKind::Progress` keep-alives every 60s. Progress acks preserve the full JetStream redelivery budget, so a parked message survives an arbitrarily long update and is processed by the *new* client after restart. The reverse guard also holds: the updater defers a client swap while a tool operation is running (30-minute staleness cutoff protects against a stuck state file).

**Migration safety.** The updater uses a fresh durable/inbox identity (`…_client-updater_client-update_consumer`) with `DeliverPolicy::New`, so old in-place clients and the new updater never race over the same consumer during rollout.

### Review hardening (11 findings from external review, all applied)

Redelivered duplicates can no longer cancel a live observation window; a failed binary replace can no longer brick the install (backup path is precomputed and restored); macOS service stop tolerates already-unloaded services and polls for real termination; crash recovery runs before any network dependency; promotion requires the service to actually be running, not just a boot marker; client downloads get a dedicated 300s timeout; ACK wait raised to 600s to match real time-to-ACK; Windows builds are statically linked (`+crt-static`).

### Tests and CI

61 unit tests in the updater crate (sidecar `*_tests.rs` per module — state machine, atomic replace, encryption round-trip, version matching, archive extraction, phase-string parity with the client's probe contract) alongside the client's 108. CI (`test.yml`) lints, builds, and tests the updater on macOS and Windows next to the client.

## Commit guide

The updater-crate commits replay the oss-tenant branch history snapshot by snapshot (each state compiled and passed review there), so the crate's evolution is reviewable commit by commit:

1. `feat(client-updater): port standalone updater service with hybrid token logic` — the service skeleton, NATS listener, token handling
2. `chore(client-updater): clippy and fmt cleanups`
3. `feat(client-updater): LKG ratchet, boot-marker verify, observation auto-rollback` — the update state machine
4. `feat(client-updater): rollback flag bypasses downgrade guard, restores from reserve`
5. `fix(client-updater): review fixes — durability, ACK semantics, blocking calls, secured-dir perms`
6. `test(client-updater): unit tests for models, platform, and services`
7. `fix(client-updater): harden update lifecycle, binary swap, and crash recovery` — the 11 external review findings
8. `style(client-updater): rustfmt` — formatting only, to satisfy this repo's `fmt --check` CI
9. `feat(client): remove in-client update machinery, park tool ops during client swap` — the client side, adapted to the agent-lib layout (the oss-tenant client commits were written against the old in-tenant client structure, so they land here as one adapted commit)
10. `ci: lint, build, and test openframe-client-updater`

Original per-commit diffs: oss-tenant PR #2180.

## Validation

- updater crate: 61 tests green, `clippy -D warnings` clean, `fmt --check` clean
- agent-lib: 108 tests green, clippy + fmt clean for both the lib and `--features bin`
- crate content verified byte-identical to the oss-tenant branch tip before formatting

## Follow-ups (not in this PR)

- saas-tenant thin `openframe-client-updater` bin + release matrix entry (mirrors the client's thin-bin pattern)
- registry JSON download links (oss-tenant PR #2254) must move from oss-tenant releases to saas-tenant releases once the updater ships from there


clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,670 @@
...
+        state.downloaded_binary_path = Some(temp_path.to_string_lossy().to_string());
+        self.state_service.save(&state)?;
+
+        self.state_service
+            .transition(&mut state, UpdaterPhase::StoppingService)?;
+        self.progress_publisher
+            .publish(&UpdaterPhase::StoppingService, version)
+            .await;
+
+        if let Err(e) = ServiceManagerService::stop_async(CLIENT_SERVICE_FULL_NAME).await {
```
denys-gif, line 268: missing reverse guard - any_tool_op_in_progress() is not used in the client. A CLIENT_UPDATE landing mid tool-install SIGTERMs the client, leaving the record Installing, the directory half-populated and orphaned children

clients/openframe-client-updater/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+pub mod clients;
+pub mod config;
```
denys-gif, line 2: service_adapter, platform/directories, platform/permissions, encryption_service, nats_connection_manager, github_download_service, utils/jwt are copies of client code in the same repo.

Every cross-crate contract - AES key, secured/ layout, LKG artifacts, phase strings is duplicated by convention only.

we should consider creating `openframe-client-common` crate for common code

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,670 @@
...
+            if let Err(e) = ServiceManagerService::stop_async(CLIENT_SERVICE_FULL_NAME).await {
+                warn!("Failed to stop service before rollback restore: {:#}", e);
+            }
+        }
+
+        // Prefer the last-known-good reserve over the pre-swap backup: the
+        // backup is merely what ran before, the reserve is what verified.
+        // The reserve is copied (it must survive); the backup is renamed.
+        let reserve = self.lkg_service.reserve_path();
+        let use_reserve = reserve.exists();
```
denys-gif: use_reserve = reserve.exists() picks the reserve without checking its version differs from the target that just failed, then line 569 deletes the pre-swap backup. A failed rollback:true restores the same non-booting binary and destroys the only working copy. 

Only prefer the reserve when its anchor version differs from the failed target, and keep the backup until the restored binary has booted.

clients/openframe-client-updater/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    let version =
```
denys-gif: please make sure the version here is correctly handled by CI so there would be no cases where we push 0.1.0 to prod

clients/openframe-client-updater/src/services/updater_orchestrator.rs
```diff
@@ -0,0 +1,423 @@
...
+                            verified = true;
+                            break;
+                        }
+                        tokio::time::sleep(Duration::from_secs(BOOT_MARKER_POLL_INTERVAL_SECS))
+                            .await;
+                        elapsed += BOOT_MARKER_POLL_INTERVAL_SECS;
+                    }
+                }
+
+                let running = matches!(
```
denys-gif: matches!(is_running(...), Ok(true)) collapses Err into false. A transient SCM or launchctl failure right after a reboot rolls back a perfectly healthy update.

Treat Err as unknown and retry the query a few times before deciding to roll back.

clients/openframe-client/src/platform/client_update_probe.rs
```diff
@@ -0,0 +1,36 @@
+use crate::platform::directories::get_secured_directory;
+
+pub const UPDATER_TOOL_AGENT_ID: &str = "openframe-client-updater";
+
+const UPDATER_STATE_FILE_NAME: &str = "updater_state.json";
+const UPDATER_STATE_STALE_SECS: u64 = 30 * 60;
+const IN_FLIGHT_UPDATER_PHASES: [&str; 8] = [
```
denys-gif, line 7: idle is missing though it is persisted on the restore-from-reserve path, opening a window where a tool op dispatches right before the swap; observing is included, parking every tool operation for 600s after the client is already healthy.

Add idle to the in-flight list and drop observing from it.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,670 @@
...
+            state_service,
+            progress_publisher,
+            lkg_service,
+            in_progress: Arc::new(Mutex::new(false)),
+            generation: Arc::new(AtomicU64::new(0)),
+        }
+    }
+
+    pub async fn process_update(&self, msg: ClientUpdateMessage) -> Result<()> {
+        let mut guard = self.in_progress.lock().await;
```
denys-gif: in_progress is not RAII. the flag is set, the guard dropped, and only reset on the normal return path. Any panic inside run_update leaves it stuck true and the machine can never be updated again.

Hold an OwnedMutexGuard across run_update so the slot is released on unwind.

clients/openframe-client-updater/src/platform/atomic_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+        )
+    })?;
+
+    info!("Restored backup to: {}", target.display());
+    Ok(())
+}
+
+/// Restores `source` to `target` by copy, leaving `source` in place.
+/// Used for the last-known-good reserve, which must survive the rollback.
+pub fn restore_copy(source: &Path, target: &Path) -> Result<()> {
```
denys-gif, line 54: restore/restore_copy remove the target first and only then rename/copy in, unlike the careful rename - based replace(). A partial copy leaves a truncated binary, and after three failed attempts there is no client binary at all.

write to a temp sibling and rename into place, reusing write_temp.

clients/openframe-client-updater/src/config/updater_config.rs
```diff
@@ -0,0 +1,72 @@
...
+pub const DOWNLOAD_CLIENT_TIMEOUT_SECS: u64 = 300;
+pub const MIN_BINARY_SIZE_BYTES: u64 = 100 * 1024; // 100 KB
+
+// NATS consumer settings
+pub const CLIENT_UPDATE_STREAM: &str = "CLIENT_UPDATE";
+pub const CLIENT_UPDATE_FILTER_SUBJECT: &str = "machine.all.client-update";
+/// Must comfortably exceed a normal update's time-to-ACK (download + service
+/// stop + swap + up to 90s boot wait) — at 120s the same message was
+/// redelivered mid-update and killed the observation window right after boot.
+pub const CLIENT_UPDATE_ACK_WAIT_SECS: u64 = 600;
```
denys-gif, line 21: CLIENT_UPDATE_ACK_WAIT_SECS=600 but download_with_retry can burn 3×300s plus backoff = 906s. Redelivery fires mid-download and the self-inflicted retries exhaust max_deliver=10.

should consider to cap the total download budget below ack_wait.

clients/openframe-client-updater/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+    let log_path = log_file_path(dir_manager);
+
+    // Ensure the tool folder exists before opening the file.
+    if let Some(parent) = log_path.parent() {
+        if let Err(e) = std::fs::create_dir_all(parent) {
+            eprintln!("Failed to create log directory {}: {}", parent.display(), e);
+        }
+    }
+
+    let file = std::fs::OpenOptions::new()
```
denys-gif: updater log never rotated. should add log rotation for updater log

clients/openframe-client-updater/src/services/updater_state_service.rs
```diff
@@ -0,0 +1,103 @@
...
+        if let Some(parent) = self.state_file_path.parent() {
+            fs::create_dir_all(parent)
+                .with_context(|| format!("Failed to create dir {}", parent.display()))?;
+        }
+
+        let json =
+            serde_json::to_string_pretty(state).context("Failed to serialize updater state")?;
+
+        let temp_path = self.state_file_path.with_extension("json.tmp");
+        fs::write(&temp_path, json)
```
denys-gif: updater_state.json is the sole input to crash recovery, yet save() never syncs the file or its parent, while write_temp syncs both for the binary. A power loss right after a phase transition loses the state and strands a half-swapped binary

sync_all() the temp file and the parent directory before and after the rename.

clients/openframe-client-updater/src/services/updater_orchestrator.rs
```diff
@@ -0,0 +1,423 @@
...
+            // The boot marker was cleared before StartingService, so in these
+            // phases the marker is the source of truth: only a marker carrying
+            // the target version proves the new binary booted. The marker alone
+            // proves a *past* boot though — promotion additionally requires the
+            // service to be running now (started here if needed).
+            UpdaterPhase::StartingService
+            | UpdaterPhase::VerifyingBoot
+            | UpdaterPhase::Observing => {
+                if !matches!(
+                    ServiceManagerService::is_running(CLIENT_SERVICE_FULL_NAME),
```
denys-gif: Crash recovery calls the sync is_running/start and a ~20MB fs::copy inside an async fn, even though *_async spawn_blocking wrappers exist for exactly this; archive extraction is synchronous too.

Use the existing *_async variants and move extraction and file copies into spawn_blocking.

clients/openframe-client-updater/src/services/agent_configuration_service.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub async fn get_machine_id(&self) -> Result<String> {
+        Ok(self.read_config()?.machine_id)
+    }
+
+    pub async fn get_client_credentials(&self) -> Result<(String, String)> {
+        let cfg = self.read_config()?;
+        Ok((cfg.client_id, cfg.client_secret))
+    }
+
+    pub async fn get_access_token(&self) -> Result<String> {
```
denys-gif: get_access_token/get_refresh_token/update_tokens, get_initial_key, is_installed, encrypt and four constants are never called; thiserror, winreg, winapi, tokio-rustls and embed-resource are never referenced. Nothing warns because everything is pub in a lib crate.

Should remove the unused items and dependencies before merge

clients/openframe-client/src/listener/tool_installation_message_listener.rs
```diff
@@ -158,6 +156,13 @@ impl ToolInstallationMessageListener {
     async fn dispatch(&self, message: Message, tool_installation_message: ToolInstallationMessage) {
         let tool_agent_id = tool_installation_message.tool_agent_id.clone();
 
+        if tool_agent_id == UPDATER_TOOL_AGENT_ID {
```
denys-gif: Please make sure we are not overlaping guards. The same in_flight_client_update_phase() check runs at three layers with three different semantics (park / silent return / bail!). The inner two are unreachable except in a microsecond race, and when they do fire they drop the message with no ack, burning a delivery attempt the parking design exists to preserve.

clients/openframe-client-updater/src/services/client_update_service.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. The updater is a separate process, so it cannot call `any_tool_op_in_progress()`; the client now mirrors its in-flight tool ops to `{secured}/tool_ops_in_flight.json` from `ToolRunManager::mark_updating` / `clear_updating` (JSON array of tool ids, removed when empty and at client startup). The updater's `ClientToolOpProbe` waits up to 60s (`TOOL_OP_QUIESCE_WAIT_SECS`) for it to drain right before `StoppingService`; if ops are still running it removes the temp binary, clears its state and returns `Err` so the message is redelivered. No deadlock with the client-side park: new tool ops are parked while `updater_state.json` shows an in-flight phase, so only ops that started before the update hold the marker. A marker older than 30 min is ignored (client died mid-op).

clients/openframe-client-updater/src/lib.rs
danylo-babenko-flamingo: Agreed, but this will be done as a separate task, not in this PR (ticket: `possible_tickets/Working/CLIENT_COMMON_CRATE`). The copies have diverged too far for a file move: `directories.rs` 742 differing lines, `permissions.rs` 468, `service_adapter.rs` 660, `github_download_service.rs` 352; only `jwt.rs` is identical. Extracting `openframe-client-common` means reconciling those first, and it also has to cover the chat and desktop clients, which I'm designing as one migration. The contracts listed here (AES key, secured layout, LKG artifacts, phase strings, plus the new tool-op marker) are the crate's first contents.

clients/openframe-client-updater/src/services/client_update_service.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `restore_candidates()` takes the reserve only when the anchor version differs from the failed target (`versions_match`); otherwise the pre-swap backup comes first. After the restore the updater clears the boot marker, starts the service and waits for a boot marker (`BOOT_MARKER_WAIT_SECS`); the backup is deleted only once the reserve-restored binary has booted. If it does not boot and a backup exists, the backup is restored (rename) and the service started again. Each candidate keeps the 3 retry attempts.

clients/openframe-client-updater/build.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `build.rs` now panics when `OPENFRAME_UPDATER_VERSION` is unset (same as the client's `forward_required`), so a release job that forgets the variable fails to build instead of shipping 0.1.0. The Makefile passes `OPENFRAME_UPDATER_VERSION=$(OPENFRAME_VERSION)` with a `0.0.0-dev` default for CI lint/test. Also found that clap's `--version` printed `CARGO_PKG_VERSION` (0.1.0); it now uses `UPDATER_VERSION`. Note: no pipeline builds or releases the updater binary yet — when the saas-tenant release job is added it must pass the version the way it does for the client.

clients/openframe-client-updater/src/services/updater_orchestrator.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. New `ServiceManagerService::query_running` retries the state query 3 times (`SERVICE_STATE_QUERY_ATTEMPTS`, 2s apart) and returns `Option<bool>` where `None` = unknown. Crash recovery rolls back only when the marker is not verified or the state is `Some(false)`; an unknown state with a matching boot marker is treated as healthy (logged). The pre-check uses the same helper and only attempts a start when the state is not confirmed running.

clients/openframe-client/src/platform/client_update_probe.rs
danylo-babenko-flamingo: Fixed in f79b50fdf: `idle` added, `observing` removed from `IN_FLIGHT_UPDATER_PHASES`. An updater self-update during observation is handled by crash recovery on the next updater start (Observing arm: marker matches + running → promote).

clients/openframe-client-updater/src/services/client_update_service.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `in_progress` is now `Arc<Mutex<()>>` taken with `try_lock_owned()`; the `OwnedMutexGuard` lives across `run_update` (and across an observation rollback), so the slot is released on every exit path including unwind.

clients/openframe-client-updater/src/platform/atomic_replace.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `restore()` is a single `rename_with_retry(backup → target)` — rename replaces on both Windows and unix, the target is never removed first. `restore_copy()` reads the reserve, goes through `write_temp` (synced temp sibling next to the target, 0755 on unix) and renames it into place; a failed rename removes the temp. Existing tests (`restore_consumes_the_backup`, `restore_copy_preserves_the_source`) still pass.

clients/openframe-client-updater/src/config/updater_config.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `download_with_retry` runs under a total deadline `DOWNLOAD_TOTAL_BUDGET_SECS = 360` (per-attempt `DOWNLOAD_TIMEOUT_SECS = 180`; backoff and the jsDelivr fallback count against the same budget, and the loop stops early when the next attempt or backoff would cross it). Worst-case time-to-ACK ≈ 360 + 60 (tool-op quiesce) + 30 (stop) + ~10 (swap) + 3 + 90 (boot marker) = 553s < 600s ack_wait; the constant's doc comment spells that out.

clients/openframe-client-updater/src/main.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `utils/rotating_log_file.rs`: `RotatingLogFile` feeds the tracing non-blocking writer; at `UPDATER_LOG_MAX_BYTES` (10 MB) it closes the file, shifts `.log.1 → .log.2 → .log.3` (`UPDATER_LOG_KEEP_ROTATED = 3`), renames `.log → .log.1` and reopens the same path, so the client keeps tailing a fixed name (its `FileLogSource` resets the offset when the file shrinks). A failed rotation keeps writing and retries after another 10 MB. Unit-tested (rotates, keeps N, resumes size accounting from an existing file).

clients/openframe-client-updater/src/services/updater_state_service.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `save()` writes the temp file through `File` + `sync_all()`, renames, then `sync_all()` on the parent directory (unix; no-op on Windows where the rename is journaled); `clear()` syncs the parent after the remove.

clients/openframe-client-updater/src/services/updater_orchestrator.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. Crash recovery uses `query_running` / `start_async`; `restore_and_start` is async and runs `restore` / `restore_copy` through `utils::blocking::run_blocking` (`spawn_blocking`); `promote` (the ~20 MB copy) runs there too, in recovery and in observation. Archive extraction in `download_and_extract`, `write_temp`, `replace`, the reserve read and the rollback restores are on the blocking pool as well.

clients/openframe-client-updater/src/services/agent_configuration_service.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. Removed `get_access_token` / `get_refresh_token` / `update_tokens` (and the token fields), `get_initial_key`, `UpdaterService::is_installed`, the four unused constants (`CLIENT_UPDATE_FILTER_SUBJECT`, `SERVICE_START_TIMEOUT_SECS`, `SUBJECT_UPDATE_PROGRESS`, `SUBJECT_INSTALLED_AGENT`) and the dependencies thiserror, winreg, winapi, tokio-rustls, embed-resource (Cargo.lock pruned, `--locked` release build verified). `encrypt()` stays under `#[cfg(test)]` for the round-trip test, since production only decrypts the client's `shared_token.enc`.

clients/openframe-client/src/listener/tool_installation_message_listener.rs
danylo-babenko-flamingo: Fixed in f79b50fdf. `park_or_dispatch` in `client_update_gate` is now the only guard: the updater-targeted early returns in the installation / restart / uninstall listeners' `dispatch` and the `bail!` in `ToolAgentUpdateService::process_update` are removed, so a message that passes the park is always processed and acked or left for redelivery normally.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+        warn!("Rolling back update to v{}: {}", version, reason);
+
+        self.transition_best_effort(state, UpdaterPhase::RollingBack);
+        self.progress_publisher
+            .publish(&UpdaterPhase::RollingBack, version)
+            .await;
+
+        // The service may still be running the bad binary (e.g. it booted but
+        // reported the wrong version) — stop it before touching the file.
+        self.stop_service_if_running("before rollback restore")
```
denys-gif, line 576: The tool-op quiesce guard was added only to the forward path at line 247; rollback() goes straight to stop_service_if_running. Since observing was simultaneously removed from the client's park list, tool ops run freely for 600s and an observation rollback kills the client mid-install.

Route every client stop through one helper that performs the quiesce wait first, including rollback() and crash recovery.

clients/openframe-client-updater/src/services/updater_orchestrator.rs
```diff
@@ -0,0 +1,434 @@
...
+        let reserve = lkg_service.reserve_path().to_path_buf();
+        let reserve = reserve.exists().then_some(reserve);
+
+        // Restore sources in preference order. The reserve is copied (it must
+        // survive), the backup is renamed (consumed). After a failed boot the
+        // reserve is the stronger choice — it verified; the backup merely ran.
+        // Before the swap the backup is authoritative, and the reserve is only
+        // a last resort when the client binary itself is gone.
+        let mut candidates: Vec<(PathBuf, bool)> = Vec::new();
+        if prefer_reserve {
```
denys-gif, line 360: restore_and_start keeps the unconditional prefer_reserve ordering while restore_candidates (line 639) was fixed to skip the reserve when the anchor holds the version that just failed. Two rollback implementations, only one of them fixed.

 Delete restore_and_start's candidate logic and call restore_previous_binary so the anchor-version guard applies everywhere.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+        match ServiceManagerService::is_running_async(CLIENT_SERVICE_FULL_NAME).await {
+            Ok(true) => {}
+            Ok(false) => {
+                let reason = "Service is not running after start".to_string();
+                error!("{}", reason);
+                return self
+                    .rollback(&mut state, &target, &backup_path, version, &reason)
+                    .await;
+            }
+            Err(e) => {
```
denys-gif, line 343: query_running was introduced so a transient service-state query error is not read as "stopped", but it was wired into the orchestrator only. The Err arm here rolls back a client that started fine.

consider to use query_running and roll back only on Some(false).

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+                        ))
+                        .await;
+                    }
+                }
+            }
+        }
+        None
+    }
+
+    async fn stop_service_if_running(&self, ctx: &str) {
```
denys-gif, line 712: if let Ok(true) collapses Err into "not running", so a failed probe silently skips the stop and the rollback restores the binary while the bad client is still live. The machine keeps running the broken version while the updater reports RolledBack.

consider to use query_running and stop anyway on None - stopping an already stopped service is a no-op on both platforms.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+                }
+            }
+        };
+        state.downloaded_binary_path = Some(temp_path.to_string_lossy().to_string());
+        self.state_service.save(&state)?;
+
+        // The client parks new tool operations while this state file shows an
+        // in-flight phase; operations that started earlier must finish before the
+        // client is stopped, or their tool record and directory are left half-done.
+        if let Err(busy) = self.tool_op_probe.wait_for_quiescence().await {
```
denys-gif, line 247: A 60s quiesce timeout discards up to 360s of download and defers to a redelivery that repeats it. On a machine with regular tool activity the update never lands: ten deliveries, ten downloads, then max_deliver is gone.

consider to move wait_for_quiescence to just after the Downloading transition, which already parks new tool ops.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+        self.progress_publisher
+            .publish(&UpdaterPhase::ReplacingBinary, version)
+            .await;
+
+        let replaced = {
+            let (target, temp_path, backup_path) =
+                (target.clone(), temp_path.clone(), backup_path.clone());
+            run_blocking(move || atomic_replace::replace(&target, &temp_path, &backup_path)).await
+        };
+        if let Err(e) = replaced {
```
denys-gif, line 293: consider cleanup_temp on that branch and sweep stale update siblings at startup.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+            return Err(anyhow!(reason));
+        }
+
+        self.state_service
+            .transition(&mut state, UpdaterPhase::StoppingService)?;
+        self.progress_publisher
+            .publish(&UpdaterPhase::StoppingService, version)
+            .await;
+
+        if let Err(e) = ServiceManagerService::stop_async(CLIENT_SERVICE_FULL_NAME).await {
```
denys-gif, line 268: check in_flight_client_update_phase() on the client before mark_updating, not only when dispatching NATS messages.

clients/openframe-client-updater/src/services/client_update_service.rs
```diff
@@ -0,0 +1,800 @@
...
+                    "update already in progress, deferring v{} to redelivery",
+                    msg.version
+                ));
+            }
+        };
+
+        self.run_update(&msg).await
+    }
+
+    async fn run_update(&self, msg: &ClientUpdateMessage) -> Result<()> {
```
denys-gif, line 83: run_update is ~300 lines in an 800-line file. split into acquire_binary / prepare_swap / swap_and_verify with one shared stop_client_for(reason), and make the state paths Option<PathBuf>.

clients/openframe-client-updater/src/services/service_manager_service.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub async fn is_running_async(service_name: &'static str) -> Result<bool> {
+        tokio::task::spawn_blocking(move || Self::is_running(service_name))
+            .await
+            .context("Service status task failed to join")?
+    }
+
+    /// Service state with transient query failures retried; `None` means the
+    /// state could not be determined (SCM/launchctl kept failing) — callers
+    /// must not treat that as "stopped".
+    pub async fn query_running(service_name: &'static str) -> Option<bool> {
```
denys-gif, line 59: is_running, is_running_async and query_running all answer "is the service up", and the three callers handle Err three different ways. make query_running the only public entry point returning Option<bool> and keep is_running private to the platform impls.

PR #1816, opened 2026-08-17
feat(notification): openframe-notification-core module — spec SPI, Notifier, attributes

semen-flamingo:
Step 1 of the notification-architecture rework (producers emit `type + facts`, the notification module owns text/recipients/classification). This PR changes **zero runtime behavior** — it relocates the pipeline into its own module and lays in the new, not-yet-called emission API plus additive data fields.

## Moved (packages now `com.openframe.notification.*`, history preserved)
- from `openframe-data-nats`: `NotificationBroadcaster`, `NotificationCommand`, `NotificationChannelDispatcher`, `NotificationChannel` (SPI), `NotificationChannelExecutorConfig`, `NotificationReadEventNatsListener` (+ tests)
- from `openframe-data-mongo-sync`: `NotificationReadStateService`, `NotificationReadEvent(+Listener)` (+ tests with own test app/base)

## New emission API (registered beans; first caller arrives with the tenant spec catalog)
- `NotificationType` — interface in the lib, an enum next to each spec catalog implements it (extensible-enum pattern); only the name crosses process boundaries
- `NotificationTypeSpec` — one class per type: declared required/optional keys, checkbox/category/severity constants, `audience()`, `compose()` + temporary `buildLegacyContext()` so documents keep their current shape through the migration. Specs are pure — producers pass **all** facts (fat payload), there is no enrichment step
- `NotificationRequest {type, attrs}` with a null-skipping builder — the single producer argument; spec-side static factories give compile-time completeness without per-type classes
- `Attrs` (required-keys validation; undeclared keys dropped with a WARN — producers may emit facts before the catalog consumes them), `Audience` (null-safe, `except()`), `NotificationText`, `NotificationTypeRegistry` (duplicate types fail at startup), `Notifier.notify(request[, correlationId])` — **never throws**: the catch-and-log every dispatcher hand-wrote lives inside now
- `AudienceResolver` (active admins) joins the module; the tenant's copy must be deleted when it switches (duplicate bean name otherwise)

## Additive fields — written as `null` today, no reader changes
- `Notification.type` + `Notification.attributes` (Mongo doc); same pair on the NATS `NotificationMessage` next to the untouched `context` (desktop client verified tolerant to extra fields)
- GraphQL: `Notification.type`, `Notification.attributes` (new `JSON` scalar, same duplicated-scalar precedent as `Instant`), `context` made nullable ahead of its eventual retirement

## For consumers on version bump
Services discover these beans via explicit package scans — add `com.openframe.notification` to the scan list (tenant PR does this, and deletes its `AudienceResolver` copy).

https://claude.ai/code/session_013Za37QM5coc8FGW6gVNC9e


openframe-notification-core/src/main/java/com/openframe/notification/Notifier.java
```diff
@@ -0,0 +1,78 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+import java.util.Map;
+import java.util.Set;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class Notifier {
```
kirill-567: My proposal not to call service classes like models.

openframe-notification-core/src/main/java/com/openframe/notification/service/NotificationCommand.java
```diff
@@ -20,6 +21,9 @@ public final class NotificationCommand {
     private final String correlationId;
     private final Set<String> adminAudience;
     private final Set<String> machineAudience;
+    // Null until the caller is a spec-driven Notifier; legacy dispatchers don't set them.
+    private final String type;
```
kirill-567: I'm agree for the map and string for the graphql. 
But for the model please try to find way to consume model + enum.
As it's "note" for the service what data it needs to send.

kirill-567: Let me know if you have concerns.

openframe-notification-core/src/main/java/com/openframe/notification/service/NotificationCommand.java
```diff
@@ -28,7 +32,9 @@ public final class NotificationCommand {
                         NotificationContext context,
                         String correlationId,
                         Set<String> adminAudience,
-                        Set<String> machineAudience) {
+                        Set<String> machineAudience,
```
kirill-567: I wonder if we can resolve audience at the notification lib level.
Then publisher can say like "For all admins", "For all machines" etc.
Does it make sence?

openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+
+import static java.util.stream.Collectors.toUnmodifiableSet;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+@Slf4j
+public final class Attrs {
```
kirill-567, line 18: A bit miss idea of this class.

kirill-567:
Let’s merge but please pay attention to the comments next pr

kirill-567 (dismissed):

openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
semen-flamingo: Class to work with attributes
validation
dropping of undeclared keys by spec
typed access 
etc

openframe-notification-core/src/main/java/com/openframe/notification/service/NotificationCommand.java
semen-flamingo: fixed

openframe-notification-core/src/main/java/com/openframe/notification/service/NotificationCommand.java
semen-flamingo: Nice idea
Now Audience is just a declarative class
audience resolver resolves it into recipients class which contains real user ids based on audience

openframe-notification-core/src/main/java/com/openframe/notification/Notifier.java
semen-flamingo: Now notification emitter

kirill-567 (approved):

semen-flamingo:
⚠️ **Before bumping `openframe.oss.libs.version` past this release** — three consumers, three different outcomes:

- **saas-lib**: compile break (loud) — `FcmPushChannel` imports the moved channel SPI. Fix = retarget one import (planned follow-up PR).
- **tenant**: compile break (loud) — the 5 notification dispatchers + IT support apps import moved classes. Fix = the planned catalog PR.
- **oss-tenant / any service**: compiles clean but **fails at startup** — `NotificationReadStateService` & co moved to `com.openframe.notification`, which is not in the services' explicit `@ComponentScan` lists. Fix = add `com.openframe.notification` to the scan (one line) before or with the bump.

Runtime contracts (Mongo/NATS/GraphQL) are additive-only — nothing to coordinate there.

PR #1817, opened 2026-08-17
Fixed: Render source in ScriptExecution

andrii-flamingo:


andrii-flamingo:
Not actual anymore. This bug is fixed as a side effect in PR https://github.com/flamingo-stack/openframe-oss-lib/pull/1831
image

PR #1818, opened 2026-08-17
[AUTH] Rejected refresh must not clear auth cookies (prod kick-out fix)

aliaska-varieva:
Root cause of the recurring "users kicked out of prod" reports.

**Mechanism** — three interlocking parts:
1. Refresh rotation (`reuseRefreshTokens=false`): every refresh invalidates the previous token, so stale copies exist by design (parallel tabs, in-flight requests, WS-reconnect refresh triggers).
2. Concurrent refreshes race; the loser presents a rotated-away token → `invalid_grant`.
3. The BFF answered that with `unauthorizedWithClearedCookies()` — wiping the winner's freshly rotated, valid cookie and destroying a healthy session.

**Prod evidence** (`oauth2_authorizations` + gateway logs, tenant `2f9fdf74…`): user logs in at `17:35:14.475`; a stale refresh from their previous session lands at `17:35:14.622` — 150ms later — and its response wipes the brand-new cookies. User is kicked immediately after logging in. Repeats ~daily across tenants.

**Change**: rejected (`invalid_grant`) or unknown refresh tokens return a plain 401 with cookies preserved. The no-token branch still clears (nothing valid to protect), and explicit `/oauth/logout` remains the only path that clears an established session.

**Follow-ups** (separate): refresh coalescing in the BFF (Redis-backed, dedupes concurrent refreshes across replicas) and/or frontend single-flight refresh — either removes the race itself; this PR removes its blast damage.

**Verification after deploy**: the log correlation that defines a kick — `invalid_grant` refresh followed within minutes by a fresh `authorization_code` login from the same tenant — should disappear; `Refresh rejected (cookies preserved)` entries may remain and are harmless.

### DIFF
--- a/openframe-security-oauth/src/main/java/com/openframe/security/oauth/controller/OAuthBffController.java
+++ b/openframe-security-oauth/src/main/java/com/openframe/security/oauth/controller/OAuthBffController.java
@@ -116,11 +116,23 @@ public class OAuthBffController {
 
         return tokensMono
                 .map(tokens -> buildNoContentWithCookies(tokens, includeHeaders))
-                .switchIfEmpty(Mono.fromSupplier(this::unauthorizedWithClearedCookies))
+                // A rejected/unknown refresh token must NOT clear cookies. With rotation
+                // (reuseRefreshTokens=false), concurrent refreshes race: the winner rotates the
+                // token and Set-Cookies the new one; the loser's stale token lands here — and
+                // wiping cookies in that response destroys the winner's freshly established
+                // session (observed in prod: a login killed 150ms after completion by a stale
+                // refresh from the previous session). A plain 401 leaves the browser's current —
+                // possibly newer and valid — cookies intact; explicit /logout remains the only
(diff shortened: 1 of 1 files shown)


aliaska-varieva:
Provenance note: the cookie-clearing this PR removes was introduced in #1610 (Aug 3) — a log-noise-reduction change whose actual goals (401 instead of 500 for expired tokens, `InvalidRefreshTokenException`, WARN logging, no upstream-body leak) are **all preserved here**. The cookie wipe was incidental to that PR and had no stated motivation; two weeks of prod data since then shows it destroying freshly-rotated sessions when concurrent refreshes race (full trace in the PR description). Behavior returns to the pre-Aug-3 semantics: plain 401, cookies untouched.

@oleksandrd-flamingo — flagging since #1610 was yours: if the cookie-clearing was covering a case not in the PR text (e.g. a stuck retry loop you observed), please shout — otherwise this should be a safe partial revert.

PR #1821, opened 2026-08-17
Added Grafana metrics for RMM

andrii-flamingo:
Related with https://github.com/flamingo-stack/openframe-saas-shared/pull/1794


openframe-stream-service-core/src/main/java/com/openframe/stream/handler/ScriptExecutionStatusUpdateHandler.java
```diff
@@ -169,6 +171,7 @@ private void applyResult(ScriptExecution row, JsonNode after) {
         row.setStderrTruncated(truncStderr.truncated);
 
         scriptExecutionRepository.save(row);
+        executionMetrics.recordCompleted(RmmExecutionMetrics.KIND_SCRIPT, newStatus, row.getDispatchedAt(), now);
```
yaroslavmokflmg, line 174: Placement is right - it sits after `save`, so a failed write cannot inflate the counter, and `applyResult` returns early when the row is no longer RUNNING, so there is no double counting with the watchdog. Both good.

But that early return has a consequence worth naming: an execution the watchdog already reaped never reaches this counter. So `openframe.rmm.execution.completed` means "results that came back", not "all executions" - and the failure ratio built on it in the dashboard PR excludes exactly the worst failure mode (nothing came back at all).

Two ways out, either is fine, but it should be a decision rather than an accident:

- emit `completed{kind, status=FAILED, reason=watchdog}` from `ScriptExecutionWatchdogService` too, and keep `watchdog.reaped` as the drill-down; or
- - leave it as is and rename the dashboard panel to "Result failure ratio".
I left the same note on flamingo-stack/openframe-saas-shared#1794 so the two PRs stay consistent.

openframe-stream-service-core/src/main/java/com/openframe/stream/metrics/RmmExecutionMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Nullable
+    private final MeterRegistry registry;
+
+    public RmmExecutionMetrics(@Nullable MeterRegistry registry) {
+        this.registry = registry;
+    }
+
+    public void recordCompleted(String kind, ExecutionStatus status, Instant dispatchedAt, Instant finishedAt) {
+        if (registry == null) {
```
yaroslavmokflmg: This null-safety is dead code. `openframe-stream-service-core` depends on `openframe-core`, which pulls in `spring-boot-starter-actuator` and `micrometer-registry-prometheus` - so a `MeterRegistry` bean always exists in every service built on this module. Same for `openframe-client-core`, which is why the `@Nullable` field and the `if (meterRegistry != null)` in `ScriptExecutionWatchdogService` are redundant too.

It is also the wrong failure mode to build in: if the actuator dependency ever moves, this silently stops emitting metrics instead of failing at startup, and the dashboard just goes quiet. I would inject `MeterRegistry` plainly and let the context fail loudly if it is ever missing.

Related: the explicit `micrometer-core` dependency added to both poms is already transitive via `openframe-core`, so those four lines can go as well.

openframe-stream-service-core/src/main/java/com/openframe/stream/metrics/RmmExecutionMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+
+        if (dispatchedAt != null && finishedAt != null) {
+            Timer.builder(LATENCY_TIMER)
+                    .description(LATENCY_DESCRIPTION)
+                    .tag(TAG_KIND, kind)
+                    .tag(TAG_STATUS, statusTag)
+                    .publishPercentileHistogram()
+                    .register(registry)
+                    .record(Duration.between(dispatchedAt, finishedAt));
+        }
```
yaroslavmokflmg, line 48: `publishPercentileHistogram()` with no `minimumExpectedValue` / `maximumExpectedValue` emits Micrometer's full default bucket ladder, and it does that per `(kind, status)` pair. Two consequences.

Series count in Mimir: this one timer is by far the most expensive thing in the PR, nothing else here comes close.

Range: dispatch to result write-back for RMM legitimately runs into minutes - the watchdog's own fallback threshold is 600s, and per row it is `timeoutSeconds + graceSeconds`. Everything above the top bucket collapses into `+Inf`, so p95/p99 on the dashboard clip instead of showing the tail you actually want to see.

Suggest bounding it to the range we care about - `.minimumExpectedValue(Duration.ofSeconds(1))` and `.maximumExpectedValue(Duration.ofMinutes(15))`, or explicit SLO buckets if there are numbers we want to alert on.

Either way, worth curling `/actuator/prometheus` on dev after the change and counting `openframe_rmm_execution_latency_seconds_bucket` lines before merging. That number is the real cost of this PR and it is cheap to check.

openframe-stream-service-core/src/test/java/com/openframe/stream/handler/CommandExecutionStatusUpdateHandlerTest.java
```diff
@@ -45,7 +47,7 @@ class CommandExecutionStatusUpdateHandlerTest {
 
     @BeforeEach
     void setUp() {
-        handler = new CommandExecutionStatusUpdateHandler(commandExecutionRepository);
+        handler = new CommandExecutionStatusUpdateHandler(commandExecutionRepository, new RmmExecutionMetrics(new SimpleMeterRegistry()));
```
yaroslavmokflmg: All four test changes are constructor plumbing - `SimpleMeterRegistry` is passed in, but nothing asserts on it. So the PR adds behaviour that no test covers.

That matters more than usual here, because the metric names are a contract with the dashboard in flamingo-stack/openframe-saas-shared#1794. Rename `openframe.rmm.execution.completed` and Grafana goes blank with every test still green.

One assertion per handler is enough - keep a reference to the registry instead of inlining it:

```java
assertEquals(1.0, registry.get("openframe.rmm.execution.completed")
        .tags("kind", "command", "status", "SUCCESS").counter().count());
```

Worth also asserting the "watchdog beat us to it" path records nothing, since that is the subtle half of the logic.

yaroslavmokflmg (commented):
No blockers here - the instrumentation points are the right ones. Increments sit after `save`/`saveAll` so a failed write cannot inflate counts, the "watchdog beat us to it" path is deliberately not counted, and tag cardinality is bounded (3 statuses x 2 kinds). Four things inline, of which only the histogram bounds really need doing before merge.

Smaller things, not worth blocking on:

Tag values are uppercase (`SUCCESS` / `FAILED`, straight from `status.name()`), while the existing `openframe_push_outbox_rows_total` uses lowercase (`sent`, `failed`). Cosmetic, but it is baked into the dashboard queries, so now is the moment to pick one.

Metric-name constants are split across modules - `"openframe.rmm.execution.watchdog.reaped"` lives in `ScriptExecutionWatchdogService`, the rest in `RmmExecutionMetrics`. A shared holder in `openframe-core` would keep the `openframe.rmm.*` namespace in one place.

The latency timer spans two pods' clocks: `dispatchedAt` is written by `openframe-api-lib` in the API pod, `finishedAt` is `Instant.now()` in the stream pod. Fine in practice, and Micrometer drops negative durations, but worth remembering when reading p99.

Two follow-ups for later, not this PR: a dispatch-side counter (`dispatched - completed` is what tells you results stopped coming back, immediately, instead of 10 minutes later when the watchdog reaps), and a watchdog for `CommandExecution` - right now `kind` on the reaped counter only ever takes the value `script`.

openframe-client-core/src/main/java/com/openframe/client/service/rmm/watchdog/ScriptExecutionWatchdogService.java
```diff
@@ -19,8 +21,14 @@ public class ScriptExecutionWatchdogService {
...
     private static final String STUCK_ERROR_MESSAGE_FORMAT = "No result received within %d seconds; watchdog marked execution as failed";
 
+    private static final String REAPED_COUNTER = "openframe.rmm.execution.watchdog.reaped";
+    private static final String TAG_KIND = "kind";
+    private static final String KIND_SCRIPT = "script";
+
     private final ScriptExecutionRepository scriptExecutionRepository;
     private final ScheduleJobExecutionWatchdogService headerWatchdogService;
+    @Nullable
+    private final MeterRegistry meterRegistry;
```
kirill-567: Why not create MetricPublisher and incapsulate this logic there?
Without storing all this constants here.

openframe-stream-service-core/src/test/java/com/openframe/stream/handler/CommandExecutionStatusUpdateHandlerTest.java
andrii-flamingo: Fixed — both tests now assert the execution.completed counter (kind/status) on success and that the already-terminal path records nothing.

openframe-client-core/src/main/java/com/openframe/client/service/rmm/watchdog/ScriptExecutionWatchdogService.java
andrii-flamingo: Good catch. Agree. Updated

yaroslavmokflmg (approved):

kirill-567 (approved):

PR #1829, opened 2026-08-18
feat(client): pass userId through install, registration, and the reinstall store

danylo-babenko-flamingo:
Feature:

Pass the installing user's identity through the device installation command so the backend can associate a registered device with the user who initiated the installation. The UI-generated install command gains an optional --userId parameter:

openframe-client install --serverUrl <url> --initialKey <key> --orgId <org> --userId <user>

The client stores the value, sends it to the registration endpoint, and keeps it across uninstall/reinstall cycles — so a device re-registering later still carries the original installer's identity. The parameter is fully optional: machines installed without it (the entire existing fleet) behave exactly as today, with no new validation and no blockers anywhere in the install, registration, or reinstall paths.

Solution:

userId follows the same path orgId already takes, plus one extra stop in the reinstall-surviving store:

- CLI → config: new optional --userId argument on install (cli.rs), carried via InstallConfigParams into InitialConfiguration and persisted in {secured}/initial_config.json. The field is #[serde(default)], so configs written before this change keep parsing. No validation — unlike serverUrl/initialKey/orgId, an absent userId is a normal state.
- Registration: AgentRegistrationRequest gains user_id, serialized as camelCase userId and skipped entirely when empty (same pattern as organizationId), sent on both /register and /reinstall.
- Reinstall survival: after a successful registration the effective userId is written into PersistedMachineInfo — the store that already survives uninstall/reinstall (Windows: HKLM\SOFTWARE\OpenFrame registry record; macOS: /Library/Preferences/com.openframe.client.plist; Linux: /etc/openframe/machine.json). resolve_user_id decides the effective value: an install-time --userId always wins; a reinstall without one falls back to the persisted value, so the association is not lost when the command omits it.
- Backward compatibility: ensure_complete still validates only machine_id/client_secret, so persisted records written before the field existed load unchanged. The other PersistedMachineInfo writers (legacy credential backfill in service.rs, the deregistration probe) explicitly persist no userId.

Validation: cargo fmt --check + clippy -D warnings clean; 132 tests passing, including new coverage for camelCase serialization, omit-when-empty, persistence round-trip, and legacy-record compatibility.


clients/openframe-client/src/models/agent_registration_request.rs
```diff
@@ -9,7 +9,37 @@ pub struct AgentRegistrationRequest {
...
     pub organization_id: String,
+    #[serde(skip_serializing_if = "String::is_empty", default)]
+    pub user_id: String,
     pub os_type: String,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub tags: Vec<DeviceTag>,
 }
+
+#[cfg(test)]
+mod tests {
```
denys-gif: please add separate files for test
ex: clients/openframe-client/src/models/agent_registration_request_tests.rs

clients/openframe-client/src/platform/machine_info_persistence.rs
```diff
@@ -386,3 +390,26 @@ fn repair_permissions() -> Result<()> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn legacy_record_without_user_id_deserializes() {
+        let info = deserialize(r#"{"machine_id":"m1","client_secret":"s1"}"#).unwrap();
+        assert_eq!(info.machine_id, "m1");
+        assert!(info.user_id.is_none());
```
denys-gif: please add separate files for test
ex: clients/openframe-client/src/models/agent_registration_request_tests.rs

danylo-babenko-flamingo: Done. Thanks.

clients/openframe-client/src/models/agent_registration_request.rs
danylo-babenko-flamingo: Done.

mikhailm-coder (approved):

PR #1832, opened 2026-08-19
feat(client): report current hostname on startup for backend sync

danylo-babenko-flamingo:
Once per client startup, the agent publishes its current hostname to the per-machine subject machine.{machineId}.hostname with payload {"hostname": "<current>"}, so a renamed device resyncs with the platform without reinstalling.

- HostnameReportPublisher (new service): resolves the hostname via the existing DeviceDataFetcher (kernel hostname + macOS scutil fallbacks; skips quietly if unresolvable), reads machineId from the agent config, and publishes over the shared NatsMessagePublisher. Called once from Client::start() right after the NATS connection is up.
- Stateless and fire-and-forget: no persisted last-reported value, no old/new comparison, no timer, no reconnect hooks — the backend owns the diff against machines.hostname. A failed publish just logs a warning; the next startup retries naturally. Plain core-NATS, no JetStream.
- Trigger coverage: client startup fires on machine boot, service restart, and every client self-update — which also matches when an OS rename actually takes effect (reboot). A rename-while-running syncs at the next restart, by design.
- HostnameReportMessage (new model) with a serialization test; subject segment is a single constant if BE wants a different name.


clients/openframe-client/src/models/hostname_report_message.rs
```diff
@@ -0,0 +1,20 @@
+use serde::Serialize;
+
+#[derive(Debug, Serialize)]
+pub struct HostnameReportMessage {
+    pub hostname: String,
+}
+
+#[cfg(test)]
```
denys-gif: the test is not needed and not worth creating a separate file for it

danylo-babenko-flamingo: Done.

PR #1837, opened 2026-08-19
feat(video-bites): featured_at + featured-recency comparator + VideoBitesStrip sortComparator

michaelassraf:
## What

- `VideoTeaser.featured_at` — ISO timestamp stamped when an admin features a bite (cleared on unfeature). Keeps the type's "true superset of persisted fields" contract accurate.
- `sortBitesByFeaturedAtDesc` in `video-bites-shared.ts` — featured-recency comparator (`featured_at` with `created_at` fallback for legacy bites, missing dates last).
- `VideoBitesStrip.sortComparator` prop (default unchanged: `sortBitesByCreatedAtDesc`) so featured surfaces can display in the same order their server aggregation is capped by.
- Vitest coverage for both comparators, including the regression case below.

## Why

Featured surfaces (flamingo/openframe homepage "Key Moments" strip) are capped at N and were ranked by the bite's original `created_at`, so featuring a bite on an older entity was a **silent no-op** once N newer featured bites existed — the 2026-08-18 "highlight marked in the modal but missing from the homepage" incident. Ranking by featuring time makes the star do what it promises.

## Reviewer notes

- Companion hub PR (stamps `featured_at` in the bites editor, ranks the aggregation, seed-then-hydrate strip): flamingo-stack/multi-platform-hub#same-branch.
- **Sequencing**: this must merge and release first; the hub PR then bumps its `openframe-frontend-core` pin (it imports the new comparator).
- No behavior change for existing strip consumers — the new prop defaults to the current comparator.

### DIFF
--- a/openframe-frontend-core/src/types/video-processing.ts
+++ b/openframe-frontend-core/src/types/video-processing.ts
@@ -19,7 +19,7 @@ export interface VideoTeaser {
   thumbnail_url?: string // Optional thumbnail image URL for video preview. If not provided, video player will show first frame automatically.
   published?: boolean // Controls visibility on public preview page (default: false, admin must select)
   featured?: boolean // Opts the bite into cross-entity featured pulls (homepage/configured surfaces). Independent of `published`.
-  featured_at?: string // ISO timestamp stamped when an admin features the bite (cleared on unfeature). Featured surfaces rank by THIS (fallback created_at) — ranking by created_at alone made featuring an old bite invisible under the surface cap.
+  featured_at?: string // ISO timestamp stamped when an admin features the bite (cleared on unfeature; all featured bites backfilled 2026-08-19). Featured surfaces rank by THIS — ranking by created_at made featuring an old bite invisible under the surface cap.
   source?: 'manual' | 'ai_generated' // Track origin of teaser
   created_at?: string // ISO timestamp for sorting (newer items first)
   start_time_ms?: number // Vizard extraction: clip start offset in the source video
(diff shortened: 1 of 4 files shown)


openframe-frontend-core/src/types/video-processing.ts
```diff
@@ -19,6 +19,7 @@ export interface VideoTeaser {
   thumbnail_url?: string // Optional thumbnail image URL for video preview. If not provided, video player will show first frame automatically.
   published?: boolean // Controls visibility on public preview page (default: false, admin must select)
   featured?: boolean // Opts the bite into cross-entity featured pulls (homepage/configured surfaces). Independent of `published`.
+  featured_at?: string // ISO timestamp stamped when an admin features the bite (cleared on unfeature). Featured surfaces rank by THIS (fallback created_at) — ranking by created_at alone made featuring an old bite invisible under the surface cap.
```
michaelassraf: why is featured at exists on video processing entity? that should be on the connection table that stores highlights

michaelassraf: Bites have no rows a connection table could reference — they're jsonb array elements inside each entity's `video_bites`/`teasers` column, and per-bite curation state already lives on the element itself (`featured`, `published`, `created_at`; VideoTeaser is documented as the superset of every field persisted in those columns). `featured_at` only timestamps the existing `featured` flag at the same altitude. The connection table story is unchanged: ENTITY-level featuring stays on `entity_platforms.is_featured`. Normalizing bites out of jsonb into their own table (which would let a junction hold this) touches the Mux pipeline, the reconcile write guards, and the RAG mappers — a separate migration effort if we ever want per-platform bite featuring; happy to spec it, but it shouldn't block this fix.

PR #1842, opened 2026-08-20
feat(notification-core): typed notification seeds — a seed class per spec replaces the raw attrs map

semen-flamingo:
## What

Reworks the emission side of the notification SPI (decisions by Semen + Kirill, 2026-08-19/20): producers no longer fill a raw `Map<String,String>` — they emit a **typed seed** (naming picked over "context" to avoid colliding with the legacy stored-context classes).

- **New `spec/NotificationSeed`** — interface with `type()`. Each spec will declare its own seed class (nested in the spec file, PR-3 in tenant). A seed arrives **self-contained**: the producer fills it from entities it already holds at the emitting moment; the seed class is the compile-checked contract of what the notification needs.
- **`NotificationRequest`** — now a single field: `seed` (extensible later without touching call sites). The attr-builder dies.
- **`NotificationTypeSpec<S extends NotificationSeed>`** — gains `getSeedClass()` and `attrs(S) → Attrs`, a **pure** seed → stored-attributes mapping (no I/O in specs — a fetch would re-read what the producer already held; keeps specs repository-free and the legacy-vs-new comparison tests trivial). `getRequiredKeys()/getOptionalKeys()` are gone: a field present in the per-type seed class *is* required, at compile time.
- **`Attrs`** — `validated()` and the declared-keys machinery die with the raw map; `of/get/optional/has/json/with/asMap` stay.
- **`NotificationEmitter`** — `seed.type()` → spec lookup → `getSeedClass().cast(seed)` (the wiring check: a seed routed to a spec built for another class fails loud) → `attrs` → audience/compose → command. Still never throws.
- **CI fix ridden along**: `openframe-notification-core` was missing from the java change filter in `changes.yaml`, so PRs touching only this module (e.g. #1834) skipped Test Java entirely. One line added.

## What this is NOT

Not a return of the legacy typed contexts: those are the **stored/wire** format (Mongo + NATS + GraphQL) and are still on their deletion path. Seeds live only at the emission call site and die inside the spec — stored shape stays `type` + `attributes` map.

## Blast radius

Zero consumers: verified no usages of `NotificationEmitter`/`NotificationTypeSpec`/`NotificationRequest` outside `openframe-notification-core` on oss-lib, saas-lib and tenant mains; #2552 doesn't touch the SPI either. Only this module's own tests changed.

## Tests

`openframe-notification-core` suite green locally on JDK 21: 78 run, 0 failed (13 skipped = Mongo ITs, failsafe-only). Emitter test proves the self-contained seed maps to the stored snapshot; swallow test covers both an unregistered type and a seed-class mismatch.

https://claude.ai/code/session_01BCDHWY4aLtcZfA2kBTpVsk

### DIFF
--- a/.github/workflows/changes.yaml
+++ b/.github/workflows/changes.yaml
@@ -59,6 +59,7 @@ jobs:
                 "openframe-gateway-service-core/**",
                 "openframe-idp-configuration/**",
                 "openframe-management-service-core/**",
+                "openframe-notification-core/**",
                 "openframe-notification-mail/**",
                 "openframe-notification-push/**",
                 "openframe-pinot-initializer/**",
--- a/openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
+++ b/openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
@@ -2,19 +2,14 @@ package com.openframe.notification.spec;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
-import java.util.Set;
-import java.util.TreeSet;
 
-import static java.util.stream.Collectors.toUnmodifiableSet;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
-@Slf4j
 public final class Attrs {
 
     private static final ObjectMapper JSON = new ObjectMapper();
@@ -23,12 +18,6 @@ public final class Attrs {
 
     private Attrs(Map<String, String> values) {
         this.values = values;
-    }
-
-    public static Attrs validated(NotificationTypeSpec spec, Map<String, String> raw) {
-        rejectMissingRequiredKeys(spec, raw);
-        Map<String, String> declared = dropUndeclaredKeys(spec, raw);
-        return new Attrs(declared);
     }
 
     public static Attrs of(Map<String, String> values) {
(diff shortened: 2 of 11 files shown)


openframe-notification-core/src/main/java/com/openframe/notification/spec/NotificationTypeRegistry.java
```diff
@@ -14,19 +14,19 @@
 @Component
 public class NotificationTypeRegistry {
 
-    private final Map<String, NotificationTypeSpec> byTypeName;
+    private final Map<String, NotificationTypeSpec<?>> byTypeName;
 
     // ObjectProvider, not List: a service with zero specs on the classpath must still boot.
     // toUnmodifiableMap throws IllegalStateException on a duplicate type — the wanted fail-fast.
-    public NotificationTypeRegistry(ObjectProvider<NotificationTypeSpec> specs) {
+    public NotificationTypeRegistry(ObjectProvider<NotificationTypeSpec<?>> specs) {
```
kirill-567, line 21: List?

openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
```diff
@@ -2,19 +2,14 @@
...
 import java.util.NoSuchElementException;
 import java.util.Optional;
-import java.util.Set;
-import java.util.TreeSet;
 
-import static java.util.stream.Collectors.toUnmodifiableSet;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
-@Slf4j
 public final class Attrs {
```
kirill-567, line 13: Do we need it as we move to the contexts?

.github/workflows/changes.yaml
```diff
@@ -59,6 +59,7 @@ jobs:
                 "openframe-gateway-service-core/**",
                 "openframe-idp-configuration/**",
                 "openframe-management-service-core/**",
+                "openframe-notification-core/**",
```
kirill-567, line 62: how it worked before?

kirill-567 (approved):

.github/workflows/changes.yaml
semen-flamingo: It just didnt work)

openframe-notification-core/src/main/java/com/openframe/notification/spec/Attrs.java
semen-flamingo: Yes because contexts only live for the notification lib user and immidiately they get converted to a map

openframe-notification-core/src/main/java/com/openframe/notification/spec/NotificationTypeRegistry.java
semen-flamingo: List will throw if no specs are in classpath
ObjectProvider is ok with that

PR #1843, opened 2026-08-20
feat(onboarding): Meet Mingo replaces Company & Team; trim the user tour enum

pavlo-flamingo:
Brings both onboarding step enums in line with the current Figma. The frontend already renders this shape (openframe-oss-frontend), running on a cast until these values exist.

## What changes

**`TenantOnboardingStep`** — `COMPANY_TEAM` → `MEET_MINGO`.
The design's fourth Initial Setup row is the Mingo intro ([Figma](https://www.figma.com/design/YQttTEQP9lQIP92VGHBwvy/openframe---dashboard?node-id=10014-85482&m=dev)); inviting teammates lives in Settings → Company & Users, not in setup.

**`UserOnboardingStep`** — drops `CUSTOMERS_SETUP` and `DEVICE_MANAGEMENT`.
Both are *tenant* Initial Setup steps, and the Get Started tour only unlocks once Initial Setup is complete — so the workspace already has a customer and a device by the time a user sees the tour. The frontend has skipped them for a while.

## Migration ships with it

`completedSteps` is a `Set<enum>`, so a document still holding `"COMPANY_TEAM"` (or the two retired user steps) cannot deserialise once the constant is gone. `PruneRetiredOnboardingStepsChangeUnit` (order `010`) `$pull`s them from `tenant_onboarding_progress` and `user_onboarding_progress` — raw collections, since the entities can no longer represent what it removes, the same approach `009` took for the identical problem. Idempotent, tenant-scoped, rollback deliberately empty.

**Deploy note:** the migration runs in the management service, the documents are read by the API service — a tenant whose management service hasn't run this yet will fail the onboarding query. The failure is graceful on the client (the onboarding chrome just doesn't render, no broken dashboard) and it only affects tenants that already used the retired steps, but it is worth making sure the migration lands everywhere.

**Effect on data:** a retired step stops counting toward "X/Y done". A tenant that finished all four Initial Setup steps but never pressed Complete drops to 3/4 with Meet Mingo outstanding — the honest state, nobody has met Mingo yet. Tenants already flagged `completed` are unaffected: separate field, and the card hides on it without reading `completedSteps`. For the user tour the visible effect is zero — those two steps were never rendered.

## Merge order

This PR, then the SDL companion flamingo-stack/openframe-saas-tenant#2688 once this is released and `openframe.oss.libs.version` is bumped there. Landing the SDL first would expose `MEET_MINGO` with no Java constant behind it.

## Not verified locally

No Maven toolchain on this machine, so nothing was compiled. CI is the first real compile.

### DIFF
--- a/openframe-management-service-core/src/main/java/com/openframe/management/migration/PruneRetiredOnboardingStepsChangeUnit.java
+++ b/openframe-management-service-core/src/main/java/com/openframe/management/migration/PruneRetiredOnboardingStepsChangeUnit.java
@@ -1,0 +1,52 @@
+package com.openframe.management.migration;
+
+import com.mongodb.client.result.UpdateResult;
+import com.openframe.data.service.TenantIdProvider;
+import io.mongock.api.annotations.ChangeUnit;
+import io.mongock.api.annotations.Execution;
+import io.mongock.api.annotations.RollbackExecution;
+import lombok.extern.slf4j.Slf4j;
+import org.bson.Document;
+import org.springframework.data.mongodb.core.MongoTemplate;
+
+import java.util.List;
+
+/**
+ * Removes steps the onboarding enums no longer declare from the stored {@code completedSteps}:
+ * {@code COMPANY_TEAM} (tenant) and {@code CUSTOMERS_SETUP} / {@code DEVICE_MANAGEMENT} (user).
+ * A document still holding one cannot deserialise into a {@code Set<enum>}, so this ships with the
+ * enum change. Raw collections for the same reason; idempotent; rollback would re-break reads.
+ */
+@Slf4j
+@ChangeUnit(id = "prune-retired-onboarding-steps", order = "010", author = "openframe")
+public class PruneRetiredOnboardingStepsChangeUnit {
+
+    private static final String TENANT_COLLECTION = "tenant_onboarding_progress";
+    private static final String USER_COLLECTION = "user_onboarding_progress";
+    private static final String TENANT_ID_FIELD = "tenantId";
+    private static final String COMPLETED_STEPS_FIELD = "completedSteps";
+
+    private static final List<String> RETIRED_TENANT_STEPS = List.of("COMPANY_TEAM");
+    private static final List<String> RETIRED_USER_STEPS = List.of("CUSTOMERS_SETUP", "DEVICE_MANAGEMENT");
+
+    @Execution
+    public void execution(MongoTemplate mongoTemplate, TenantIdProvider tenantIdProvider) {
+        String tenantId = tenantIdProvider.getTenantId();
+        prune(mongoTemplate, TENANT_COLLECTION, RETIRED_TENANT_STEPS, tenantId);
+        prune(mongoTemplate, USER_COLLECTION, RETIRED_USER_STEPS, tenantId);
+    }
+
+    @RollbackExecution
+    public void rollback() {
+    }
+
+    private void prune(MongoTemplate mongoTemplate, String collection, List<String> retiredSteps, String tenantId) {
+        UpdateResult result = mongoTemplate.getCollection(collection).updateMany(
+                new Document(TENANT_ID_FIELD, tenantId)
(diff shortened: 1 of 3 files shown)


pavlo-flamingo:
Companion SDL PR: flamingo-stack/openframe-saas-tenant#2688 — merge this one first, release, bump `openframe.oss.libs.version` there, then that one.

openframe-management-service-core/src/main/java/com/openframe/management/migration/PruneRetiredOnboardingStepsChangeUnit.java
```diff
@@ -0,0 +1,80 @@
...
+ * {@link TenantOnboardingStep} retired {@code COMPANY_TEAM} (the design replaced it with
+ * {@code MEET_MINGO}), and {@link UserOnboardingStep} retired {@code CUSTOMERS_SETUP} and
+ * {@code DEVICE_MANAGEMENT} (both are tenant Initial Setup steps; the user tour never showed them).
+ * No client renders any of the three.
+ *
+ * <p>This is the MIGRATE step of an expand/contract change: the constants are still declared (so
+ * nothing breaks whether or not this has run yet) and are dropped only in a later release, once
+ * every tenant's data has been through here. Doing the removal and this cleanup in one release
+ * would have made correctness depend on deploy order — this migration runs in the management
+ * service while the reader is the API service, and a tenant whose management service starts late,
```
semen-flamingo: Too much jdoc)

openframe-management-service-core/src/main/java/com/openframe/management/migration/PruneRetiredOnboardingStepsChangeUnit.java
```diff
@@ -0,0 +1,80 @@
...
+    /** Retired from {@link TenantOnboardingStep}. */
+    private static final List<String> RETIRED_TENANT_STEPS = List.of("COMPANY_TEAM");
+    /** Retired from {@link UserOnboardingStep}. */
+    private static final List<String> RETIRED_USER_STEPS = List.of("CUSTOMERS_SETUP", "DEVICE_MANAGEMENT");
+
+    @Execution
+    public void execution(MongoTemplate mongoTemplate, TenantIdProvider tenantIdProvider) {
+        String tenantId = tenantIdProvider.getTenantId();
+        prune(mongoTemplate, TENANT_COLLECTION, RETIRED_TENANT_STEPS, tenantId);
+        prune(mongoTemplate, USER_COLLECTION, RETIRED_USER_STEPS, tenantId);
```
semen-flamingo: I think you can remove them and not deprecate

semen-flamingo (approved):
Reviewed the enum change, the migration and every consumer of these constants across oss-lib, saas-tenant, oss-tenant, saas-lib and saas-shared — saas-tenant's SDL is the only other place they appear, and #2688 covers it.

The migration is genuinely required (`completedSteps` is a `Set<enum>`, so a stored `COMPANY_TEAM` cannot deserialise) and follows the proven shape of ChangeUnit 009: raw collection, explicit `tenantId` filter, `$in` pre-filter making it idempotent, empty rollback.

Bumped the change unit to order `011` — `010` is already taken by the companion unit in openframe-saas-tenant (`rename-ticket-label-to-tag-in-chat-documents`), and openframe-saas-management scans both from the same `com.openframe.management.migration` package, so the numbering space is shared across the two repos.

One correction to the merge order in the description: the `openframe.oss.libs.version` bump has to land **inside** #2688, not as a separate PR before it. `OnboardingService.completedTenantProgress()` returns `EnumSet.allOf(TenantOnboardingStep.class)` for super users, so the moment this lib lands in the tenant, every super-user read of `tenantOnboardingProgress` serialises `MEET_MINGO`. With the SDL still on `COMPANY_TEAM` that is a `CoercingSerializeException`, and the field is non-null, so the whole query nulls out. Landing the lib bump and the SDL together closes that window.

(Also: the description says "bump past 6.25.30" — tenant main is already on 6.25.32.)

kirill-567 (approved):

PR #1846, opened 2026-08-20
feat(meeting-scheduler): responsive design pass + the DS calendar

pavlo-flamingo:
Lib half of the onboarding "Book a call" work. App side: flamingo-stack/openframe-oss-frontend (companion PR below) — draft until both are reviewed together.

Figma: [desktop](https://www.figma.com/design/YQttTEQP9lQIP92VGHBwvy/openframe---dashboard?node-id=10014-85591&m=dev) · [tablet](https://www.figma.com/design/YQttTEQP9lQIP92VGHBwvy/openframe---dashboard?node-id=10014-84697&m=dev) · [mobile](https://www.figma.com/design/YQttTEQP9lQIP92VGHBwvy/openframe---dashboard?node-id=10014-85195&m=dev) · [confirm step](https://www.figma.com/design/YQttTEQP9lQIP92VGHBwvy/openframe---dashboard?node-id=10014-85734&m=dev)

## Layout

- **One stated height per shape** — 380 as a sidebar (`lg`), 550 as a header strip (`md`), exported as `MEETING_SCHEDULER_H` so a host can reserve the same box. Every stage fills it: skeleton, empty month, degraded, form. A `min-height` only pinned the tall stages and let the short ones resize the card under the visitor; paging a month or waiting on a fetch used to move the page.
- **Three layouts at the ODS breakpoints** (md 800 / lg 1280): stacked sections → header strip over two columns → sidebar beside two columns.
- **The calendar is a fixed 272px block**; only the times column is elastic, and it scrolls in its own box rather than growing the card. On phones the slot list gets a ~156px window of its own (as drawn) so a 40-slot day doesn't turn the card into 1500px.

## Calendar

The day layer is now the design system's own `DatePickerCalendar` instead of a second hand-styled grid:

- New `bare` variant — no card surface, no inset, for a host that already owns the panel. Fixed 272px tall at any width, matching the Figma component, so height never follows the column it lands in.
- New `disabledDays`, and controlled `month` / `onMonthChange` — paging a month here is a **data** event (the scheduler refetches availability), not just a view change.
- In `bare`, day states paint the **button** rather than the cell, so a 48×32 cell still renders a square selection instead of a stretched pill.
- **Bug fix that reaches every calendar:** today's tint and the selection both used `!important`, so on hover the winner came down to stylesheet order and a *selected today* turned grey under the pointer. The tint is now a modifier that yields to the selection — fixes the popover `DatePicker` and `DateFilterMenu` too.

## Data

`useMeetingBooking` moves to TanStack Query: availability is cached per month, so paging back is instant and the calendar stops blinking between months. **Hosts must mount a `QueryClientProvider`.**

## Form

- Fields go through `FieldWrapper`, whose message hangs out of flow — a validation error no longer grows its field, pushes everything below it down, and walks the submit button off the bottom of a card that states its height.
- Asterisks dropped in favour of `required` on the control (the design draws no asterisks; assistive tech still announces it).
- `FieldWrapper` gains `htmlFor`. It rendered a `<Label>` bound to nothing — clicking it focused nothing and AT read the field unnamed, which is why callers hand-rolled their own label instead of using the prop.

## Also

`FloatingWalkthroughVideo` gains `placement="inline"`: same card, same hover/transport/theater grammar, laid out by the host's flow instead of pinned. Overlay-only behaviours (fixed positioning, z-layer, shadow, appear delay, footer fade) drop out with it. Kept as one component — everything below the wrapper is placement-independent, and forking it is how the two would drift.

## Verification

`tsc --noEmit` and `npm run build` clean. Not visually verified by me — the consuming page needs a signed-in tenant, so every layout conclusion here came from the mocks' own numbers plus the author's screenshots.


pavlo-flamingo:
App companion: flamingo-stack/openframe-oss-frontend#262

romanivan-flamingo (approved):

